PR #11, opened 2020-02-12
Add prodigy utils to wellcomeml.prodigy

ivyleavedtoadflax:
Now that there is https://github.com/wellcometrust/deep_reference_parser, I'm moving the associated utilities out of datalabs into `wellcomeml.nlp` so we can access them easily on any platform. This is quite large, so let me know if you would prefer me to chop it up into smaller PRs, I think most of it is tests though.

The other significant change here is that I had added a `__main__.py` inspired by SpaCy. This basically gives us an entry point into the package functions like so:

```
$ python -m wellcomeml

ℹ Available commands
annotate_numbered_refs, prodigy_to_tsv, reach_to_prodigy, refs_to_token_annotations
```

This plugs into the cli of each of these functions, so you can also do:

```
$ python -m wellcomeml reach_to_prodigy --help
usage: wellcomeml reach_to_prodigy [-h] [-l 10] [-s \n] [-n] [-j ]
                                   input_file output_file

positional arguments:
  input_file            Path to jsonl file containing produced by scraper and
                        containing reference sections.
  output_file           Path to jsonl file into which prodigy format
                        references will be saved.

optional arguments:
  -h, --help            show this help message and exit
  -l 10, --lines 10     How many lines to include in an annotation example.
  -s \n, --split-char \n
                        Which character to split lines on.
  -n, --no-linebreak    Don't re-add linebreaks to the annotation examples
                        after splitting.
  -j  , --join-char     Which character should be used to join lines into an
                        annotation example.
```

I quite like this pattern for getting CLI access to the code without creating explicit entry points in `setup.py`. I think some of these functions are borderline in terms of their fit in WellcomeML, see what you think, happy to put them in `deep_reference_parser` if not here.


ivyleavedtoadflax:
Too big to review! cutting up!

PR #14, opened 2020-02-13
Add reach_to_prodigy

ivyleavedtoadflax:
Following on from #11 and #12, this adds `reach_to_prodigy` which allows easy conversion of scraped reference sections from reach into prodigy format.


ivyleavedtoadflax:
Moved to https://github.com/wellcometrust/deep_reference_parser/pull/5

PR #16, opened 2020-02-18
Add partial fit to SpacyClassifier

nsorros:
Adds partial fit to spacy classifier which allows training to take place for datasets that do not fit in memory. Spacy by default deals with this quite nicely so we are just translating it to sklearn convention.

### DIFF
--- a/tests/test_spacy_classifier.py
+++ b/tests/test_spacy_classifier.py
@@ -45,3 +45,25 @@ def test_multilabel_Y_list():
     model.fit(X, Y)
     assert model.score(X, Y) > 0.4
     assert model.predict(X).shape == (5,4)
+
+def test_partial_fit():
+    X = [
+        "One and two",
+        "One only",
+        "Three and four, nothing else",
+        "Two nothing else",
+        "Two and three"
+    ]
+    Y = [
+        [1,1,0,0],
+        [1,0,0,0],
+        [0,0,1,1],
+        [0,1,0,0],
+        [0,1,1,0]
+    ]
+
+    model = SpacyClassifier()
+    for x, y in zip(X, Y):
+        model.partial_fit([x], [y])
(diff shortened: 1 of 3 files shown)


tests/test_spacy_classifier.py
```diff
@@ -45,3 +45,25 @@ def test_multilabel_Y_list():
...
+        [1,0,0,0],
+        [0,0,1,1],
+        [0,1,0,0],
+        [0,1,1,0]
+    ]
+
+    model = SpacyClassifier()
+    for x, y in zip(X, Y):
+        model.partial_fit([x], [y])
+    assert model.score(X, Y) > 0.4
```
aCampello: We could lower this to 0.3, or even just `model.score(X, Y) > 0` to ensure the pipeline passes? I think this is trying to test whether the code runs correctly and output a non-zero number, which probably means that it is training.

wellcomeml/ml/spacy_classifier.py
```diff
@@ -132,6 +132,42 @@ def fit(self, X, Y):
...
+        """
+        Args:
+            X: 1d list of documents
+            Y: 2d numpy array (nb_examples, nb_labels)
+        TODO: Generalise to y being 1d
+        """
+        Y = np.array(Y)
+
+        if not hasattr(self, 'unique_labels'):
+            # We could also use classes here
```
aCampello: Yes. Either add a class check, or remove the parameter from above? It seems it is not being used...

nsorros: The parameter is needed to support the sklearn API

tests/test_spacy_classifier.py
nsorros: I think we need to make sure it learns something better than chance. Reducing to 0.3 for now.

PR #17, opened 2020-02-19
Add partial fit to BERT

nsorros:
Adds partial fit to BERT classifier so that we can train it in a streaming/online manner


wellcomeml/__version__.py
```diff
@@ -1,5 +1,5 @@
 __name__='wellcomeml'
-__version__='2020.2.1'
+__version__='2020.2.2'
```
ivyleavedtoadflax: Maybe don't increment the version just yet until we are ready for a new release.

PR #22, opened 2020-03-11
Optimise spacy memory consumption

nsorros:
Optimise use of memory in SpacyClassifier by
* deleting the passed data X, Y which then become X_train, Y_train, X_test, Y_test
* creating a generator that yields the transformed data that spacy needs on demand instead of materialising them in memory
* moves shuffling into X_train, Y_train since the data that spacy received are not in memory to be shuffled

### DIFF
--- a/wellcomeml/ml/spacy_classifier.py
+++ b/wellcomeml/ml/spacy_classifier.py
@@ -25,12 +25,14 @@ if is_using_gpu:
 
 class SpacyClassifier(BaseEstimator, ClassifierMixin):
     def __init__(self, threshold=0.5, n_iterations=5,
-                 batch_size=8, learning_rate=0.001, dropout=0.1):
+                 batch_size=8, learning_rate=0.001,
+                 dropout=0.1, shuffle=True):
         self.threshold = threshold
         self.batch_size = batch_size
         self.dropout = dropout
         self.learning_rate = learning_rate
         self.n_iterations=n_iterations
+        self.shuffle=shuffle
 
     def _init_nlp(self):
         self.nlp = spacy.blank('en')
@@ -98,19 +100,6 @@ class SpacyClassifier(BaseEstimator, ClassifierMixin):
                     for label in self.unique_labels
                 }
                 yield (x, {"cats": cats})
-
-        # start of problem        
-        #train_texts = X_train
-        #train_tags = self._label_binarizer_inverse_transform(Y_train)
-    
-        #train_cats = [
-        #    {
-        #        label: label in tags
-        #        for label in self.unique_labels
-        #    } for tags in train_tags
-        #]
-        #train_data = list(zip(train_texts, [{"cats": cats} for cats in train_cats]))
-        # end of problem
         
         other_pipes = [pipe for pipe in self.nlp.pipe_names if pipe != "textcat"]
         with self.nlp.disable_pipes(*other_pipes):  # only train textcat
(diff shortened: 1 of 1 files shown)


wellcomeml/ml/spacy_classifier.py
```diff
@@ -81,23 +83,24 @@ def fit(self, X, Y):
...
-                for label in self.unique_labels
-            } for tags in train_tags
-        ]
-
-        train_data = list(zip(train_texts, [{"cats": cats} for cats in train_cats]))
-
+        
+        def yield_train_data(X_train, Y_train):
+            for x, y in zip(X_train, Y_train):
+                tags = self._label_binarizer_inverse_transform([y])[0]
```
lizgzil: so I take it that this will take longer transforming them one by one, but be better for memory? Is there an option to transform Y_train in chunks?

nsorros: yes there is `self._label_binarizer_inverse_transform(Y)` but as you said this would defeat the point. additionally the code within that function is not that optimised for matrices so it makes almost no difference whether you pass the entire matrix or examples one by one. in terms of memory though it helps as you do not have to save the transformation of Y in a separate variable.

wellcomeml/ml/spacy_classifier.py
```diff
@@ -81,23 +83,24 @@ def fit(self, X, Y):
         X_train, X_test, Y_train, Y_test = train_test_split(
             X, Y, random_state=42
         )
+        # Free memory
+        del X
```
aCampello: I think there is an interesting discussion to whether `del` frees memory. Some people found it that hey needed to call the garbage collector. I think we should be good for now, but would be interesting to raise it with some engineer in the future.

wellcomeml/ml/spacy_classifier.py
```diff
@@ -81,23 +83,24 @@ def fit(self, X, Y):
...
-        ]
-
-        train_data = list(zip(train_texts, [{"cats": cats} for cats in train_cats]))
-
+        
+        def yield_train_data(X_train, Y_train):
+            for x, y in zip(X_train, Y_train):
+                tags = self._label_binarizer_inverse_transform([y])[0]
+                cats = {
+                    label: label in tags
```
aCampello: Took me a while to understand this, but nice! :D

wellcomeml/ml/spacy_classifier.py
aCampello: https://stackoverflow.com/questions/1316767/how-can-i-explicitly-free-memory-in-python - in case we still get into trouble with memory

wellcomeml/ml/spacy_classifier.py
```diff
@@ -109,8 +112,19 @@ def fit(self, X, Y):
             #dropout = decaying(0.6, 0.2, 1e-4)
             for i in range(n_iter):
                 losses = {}
-                # batch up the examples using spaCy's minibatch
-                random.shuffle(train_data)
+
+                def shuffle(X_train, Y_train):
```
aCampello: any reason why shuffle is defined here and not as a @classmethod ?

nsorros: not really, I have seen this patter in spacy here and there and using it. not sure when it is good to define functions on the fly vs not. I guess if shuffle gets reused, it would definitely move to a class method

wellcomeml/ml/spacy_classifier.py
nsorros: @jdu any thoughts on this?

nsorros:
Will merge for now as I wait to hear @jdu thoughts on it. I know from using it that it resolves the problem I had but how effective is del vs the other changes is an open question.

PR #25, opened 2020-03-25
Add exclusive classes param in SpacyClassifier

nsorros:
This PR extends SpacyClassifier to work in normal multiclass problems where the classes are exclusive essentially. This came up during firebreak but it was long time in my todo. It turned out to be slightly more complicated than I anticipated but not too much.

### DIFF
--- a/wellcomeml/ml/spacy_classifier.py
+++ b/wellcomeml/ml/spacy_classifier.py
@@ -29,6 +29,25 @@ class SpacyClassifier(BaseEstimator, ClassifierMixin):
                  batch_size=8, learning_rate=0.001,
                  dropout=0.1, shuffle=True, architecture="simple_cnn",
                  exclusive_classes=False, pre_trained_vectors_path=None):
+        """
+        Args:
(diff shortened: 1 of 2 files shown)


wellcomeml/ml/spacy_classifier.py
```diff
@@ -28,7 +28,7 @@ class SpacyClassifier(BaseEstimator, ClassifierMixin):
     def __init__(self, threshold=0.5, n_iterations=5,
                  batch_size=8, learning_rate=0.001,
                  dropout=0.1, shuffle=True, architecture="simple_cnn",
-                 pre_trained_vectors_path=None):
+                 exclusive_classes=False, pre_trained_vectors_path=None):
```
aCampello, line 31: :books: Perhaps some doctoring here will help, if you want. I had to read the PR explanation to understand what exclusive_classes does.

nsorros:
> All good to me.
> 
> 📚 If you want to add a docstring for clarifying, that would be good, but otherwise just go ahead and merge

Will add a docstring but maybe we should name different for example `multilabel=True`

PR #26, opened 2020-03-30
Add pretrain vectors

nsorros:
Adds this new feature of SpaCy https://spacy.io/usage/v2-1#pretraining.

The idea of the feature is to pretrain some vectors in an largest dataset with no labels and use those vectors in the downstream task. It should have the same effect as BERT like transfer learning.

### DIFF
--- a/wellcomeml/ml/spacy_classifier.py
+++ b/wellcomeml/ml/spacy_classifier.py
@@ -28,7 +28,7 @@ class SpacyClassifier(BaseEstimator, ClassifierMixin):
     def __init__(self, threshold=0.5, n_iterations=5,
                  batch_size=8, learning_rate=0.001,
                  dropout=0.1, shuffle=True, architecture="simple_cnn",
-                 pre_trained_vectors=None):
+                 pre_trained_vectors_path=None):
         self.threshold = threshold
         self.batch_size = batch_size
         self.dropout = dropout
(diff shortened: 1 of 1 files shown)


wellcomeml/ml/spacy_classifier.py
```diff
@@ -112,6 +114,11 @@ def yield_train_data(X_train, Y_train):
             optimizer = self.nlp.begin_training()
             optimizer.alpha = self.learning_rate
             #optimizer.L2 = 1e-4
+
+            if self.pre_trained_vectors:
```
lizgzil: might be good to name `pre_trained_vectors` so it's clear it is a path type?

wellcomeml/ml/spacy_classifier.py
```diff
@@ -112,6 +114,11 @@ def yield_train_data(X_train, Y_train):
...
+                with open(self.pre_trained_vectors, "rb") as f:
+                    self.textcat.model.tok2vec.from_bytes(f.read())
```
lizgzil: where does `textcat.model.tok2vec` get initialised if `pre_trained_vectors=None`?

wellcomeml/ml/spacy_classifier.py
nsorros: what is the name suggestion? `pre_trained_vectors_path`?

wellcomeml/ml/spacy_classifier.py
nsorros: inside spacy

wellcomeml/ml/spacy_classifier.py
lizgzil: sure!

aCampello:
LGTM too - I re-ran the pipeline and it passes now.

PR #27, opened 2020-03-31
Make tests more deterministic

nsorros:
Check whether loss is decreasing after 5 iterations and whether the model is predicting all 0 or 1.

This is to fix the tests being probabilistic by checking the performance of the model which fluctuated from run to run.


nsorros:
If this PR fails, it would be a joke.

PR #32, opened 2020-04-08
Add abstract save and load

aCampello:
- [x] Add methods to save and load embeddings in memory to the abstract class `Vectorizer`, and etc base classes  `BertVectorzier` and `WellcomeTfIdfVectorizer`

- [x] Adds tests to these classes

### DIFF
--- a/wellcomeml/ml/bert_vectorizer.py
+++ b/wellcomeml/ml/bert_vectorizer.py
@@ -35,6 +35,16 @@ class BertVectorizer(BaseEstimator, TransformerMixin):
         self.pretrained = pretrained
         self.sentence_embedding = sentence_embedding
 
+    @classmethod
+    def save_transformed(cls, path, X_transformed):
+        """Saves transformed embedded vectors"""
+        np.save(path, X_transformed)
+
+    @classmethod
+    def load_transformed(cls, path):
+        """Loads transformed embedded vectors"""
+        np.load(path)
+
     def bert_embedding(self, x):
         # Max sequence length is 512 for BERT
         if len(x) > 512:
@@ -85,7 +95,6 @@ class BertVectorizer(BaseEstimator, TransformerMixin):
         self.tokenizer = BertTokenizer.from_pretrained(model_name)
         self.model.eval()
         return self
-
 
 def _check_cache_and_download(model_name):
     """ Checks if model_name is cached and return complete path"""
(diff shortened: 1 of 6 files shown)


wellcomeml/ml/bert_vectorizer.py
```diff
@@ -35,6 +35,16 @@ def __init__(self, pretrained='bert', sentence_embedding='mean_second_to_last'):
         self.pretrained = pretrained
         self.sentence_embedding = sentence_embedding
 
+    @classmethod
+    def save_transformed(cls, path, X_transformed):
+        """Saves transformed embedded vectors"""
+        np.save(path, X_transformed)
+
+    @classmethod
+    def load_transformed(cls, path):
```
nsorros, line 44: what is the point of the above two functions? the vectorizer's job is to produce the transformed X not provide a means to save it, i think the user can do that.

on the other hand, it would be nice to provide a cache location where the vectorizer saves as it transforms and then upon subsequent transforms it checks the cache first. i would prefer the second functionality. thoughts?

aCampello: I think there is value in having an abstract "save" that will save correctly, regardless of the model/type of vectoriser, even if it seems simple, since it is very recurrent in applications.

The cache function is a good idea too. It could work like:

1. An extra parameter whether to cache the predictions or not
2. A function that indexes the inputs and tells us where to look at. This could be, for example, a way of hashing the input. Note that the hash (let's say h), should be a function of many things:

```h(input_text, pretrained, embedding_type, model_parameters)```

For example, for the tdidf vectoriser, it would also be an input of any extra parameters (`min_df`, `max_features` etc). 

3. An efficient way to build a lookup table from file. 

I am pretty sure building a hash (1-2) is no brainer, but I am not 100% sure of the best way to save/store 3. Could be a `.npy` array of mixed type, with an extra initial column

```bash
np.array([index, predictions])
```

The only problem with that is that we would need to load all predictions on file. Any thoughts?

nsorros: Saving to memory should be fine i think. The impact should be equivalent to word vectors. Alternatively we could save multiple files, one for each hash and look them up or store them in a SQLlite.

We can also save in a txt with space or tab limited values starting with hash and continuing with the embedding values, whatever is more memory friendly.

nsorros:
@aCampello will you do cache or just merge?

aCampello:
I'd like to have a go at implementing the cache strategy before.

nsorros:
> I'd like to have a go at implementing the cache strategy before.

woohoo, looking forward

aCampello:
Ok - the cache worked seemed to be bigger than I though. I will merge this one and report that work in a new branch (`feature/cache_queries`)

PR #34, opened 2020-04-17
Add CNNClassifier and KerasVectorizer

nsorros:
Adds a CNN architecture inspired from Spacy. The idea is that to develop a generic architecture that will work well for most problems so that we can use as a benchmark before we develop any custom solutions.

In particular this PR adds
* a keras vectorizer that transforms text to integers and pads. it also has the ability to create an embedding matrix given path to word vectors
* a cnn classifier that works for multilabel and multioutput cases and has a number of useful parameters such as the vocabulary size, sequence length, learning rate, dropout, epochs etc as params
* examples how to use cnn classifier
* tests for both cnn classifier and keras vectorizer


nsorros:
[Build a custom CNN in the WellcomeML library that can be used for multilabel classification](https://trello.com/c/Y6nA6rfP/960-build-a-custom-cnn-in-the-wellcomeml-library-that-can-be-used-for-multilabel-classification)

wellcomeml/ml/cnn.py
```diff
@@ -0,0 +1,68 @@
...
+                embeddings_initializer=embeddings_initializer)(inp)
+        x = tf.keras.layers.LayerNormalization()(x)
+        x = residual_conv_block(x)
+        x = residual_conv_block(x)
+        x = tf.keras.layers.GlobalMaxPooling1D()(x)
+        x = tf.keras.layers.Dense(32, activation='relu')(x)
+        x = tf.keras.layers.Dropout(self.dropout)(x)
+        x = tf.keras.layers.LayerNormalization()(x)
+
+        output_activation = 'sigmoid' if nb_outputs==1 or self.multilabel else 'softmax'
```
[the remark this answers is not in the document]
nsorros: in multi label we use sigmoid not softmax. did i misunderstand the comment?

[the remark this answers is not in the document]
ivyleavedtoadflax: I think its just terminology - I  think you are doing multiclass, not multilabel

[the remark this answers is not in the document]
nsorros: in the multilabel scenario it applies sigmoid so every neuron that outputs more than 0.5 gets assigned. am in confused?

tests/test_cnn.py
```diff
@@ -0,0 +1,43 @@
...
+        "Two and three"
+    ]
+    Y = np.array([0,0,1,1])
+
+    model = Pipeline([
+        ('vec', KerasVectorizer()),
+        ('clf', CNNClassifier())
+    ])
+    model.fit(X, Y)
+    assert model.score(X, Y) > 0.6
```
aCampello, line 20: Previously we had 0.4 for Bert classifier, and used to fail. Isn't this going to fail sometimes too?

tests/test_cnn.py
```diff
@@ -0,0 +1,43 @@
...
+        [0,0,1,1],
+        [0,1,0,0],
+        [0,1,1,0]
+    ])
+    model = Pipeline([
+        ('vec', KerasVectorizer()),
+        ('clf', CNNClassifier(multilabel=True))
+    ])
+    model.fit(X, Y)
+    assert model.score(X, Y) > 0.4
```
aCampello, line 42: Also here it is 0.4

wellcomeml/ml/cnn.py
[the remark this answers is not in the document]
ivyleavedtoadflax: That's multiclass, not multilabel. It's a very pedantic point on my part!
Multiclass (softmax) = one label applied to one of n classes
Multiclass Multilabel (sigmoid) = between 0 and n labels applied to n classes

setup.py
```diff
@@ -51,6 +51,7 @@
     ],
     extras_require={
         'deep-learning': [
+            'tensorflow',
             'torch',
             'transformers<=2.0.0',
             'spacy-transformers @ https://github.com/nsorros/spacy-transformers/tarball/master#egg=spacy-transformers',
```
aCampello, line 57: I'm a bit confused - I thought this dependency on the fork had gone away with the last merge to master. Perhaps you should rebase?

wellcomeml/ml/cnn.py
```diff
@@ -0,0 +1,68 @@
...
+Predict: softmax or sigmoid depending on number of outputs
+    and whether task is multilabel
+"""
+from sklearn.base import BaseEstimator, ClassifierMixin
+from sklearn.metrics import f1_score
+import tensorflow as tf
+
+
+class CNNClassifier(BaseEstimator, ClassifierMixin):
+    def __init__(self, context_window = 3, learning_rate=0.001, batch_size=32, nb_epochs=5, dropout=0.2, multilabel=False):
```
aCampello, line 19: extra space

wellcomeml/ml/cnn.py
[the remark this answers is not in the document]
aCampello: I think this reads ok to me - seems that @nsorros is tackling both cases:

1. If it is multilabel, you apply sigmoid to the output layer, and choose the ones with high prob (`>0.5`)
2. If not multilabel (but multiclass), you do softmax and pick the one with highest probability. 
(3 - `nb_outputs=1`). If it is a binary classification problem, softmax and sigmoid are equivalent.

wellcomeml/ml/cnn.py
```diff
@@ -0,0 +1,68 @@
...
+        nb_outputs = Y.max() if not self.multilabel else Y.shape[1]
+
+        def residual_conv_block(x1):
+            filters = x1.shape[2]
+            x2 = tf.keras.layers.Conv1D(filters, self.context_window, padding='same', activation='relu')(x1)
+            x2 = tf.keras.layers.Dropout(self.dropout)(x2)
+            x2 = tf.keras.layers.LayerNormalization()(x2)
+            return tf.keras.layers.add([x1, x2])
+
+        embeddings_initializer = tf.keras.initializers.Constant(embedding_matrix) if embedding_matrix else 'uniform'
```
aCampello, line 40: Nice - so we can initialised it with pre-trained embeddings as well?

wellcomeml/ml/cnn.py
```diff
@@ -0,0 +1,68 @@
...
+        output_activation = 'sigmoid' if nb_outputs==1 or self.multilabel else 'softmax'
+        out = tf.keras.layers.Dense(nb_outputs, activation=output_activation)(x)
+
+        self.model = tf.keras.Model(inp, out)
+        self.model.compile(optimizer=tf.keras.optimizers.Adam(lr=self.learning_rate), loss='binary_crossentropy', metrics=['accuracy'])
+        self.model.fit(X, Y, epochs=self.nb_epochs, batch_size=self.batch_size, validation_split=0.1)
+        return self
+
+    def predict(self, X):
+        return self.model(X).numpy() > 0.5
```
lizgzil, line 64: is this the case for `multilabel=True`?

tests/test_cnn.py
nsorros: This is the multilabel case where achieving more than 0.4 should be ok

tests/test_cnn.py
nsorros: I realised that i need to beat luck. A random guess would perform 0.5.

wellcomeml/ml/cnn.py
nsorros: yep, even though not documented well it is possible.

wellcomeml/ml/cnn.py
nsorros: it holds for both cases

setup.py
nsorros: I am not changing this line so whatever is in master will change. I can rebase though so that this propagates here, no prob.

wellcomeml/ml/cnn.py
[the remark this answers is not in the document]
nsorros: I am not sure if there is a disagreement :P Was the point that multilabel was multiclass multilabel, not pure multilabel? In which case I would be interested to know what is pure multilabel.

Or was it that there is not multilabel and both are consired multiclass? If so, what is multilabel then?

tests/test_cnn.py
nsorros: If we have problems we can fallback to ensuring loss is decreased

PR #40, opened 2020-04-21
Add attention

nsorros:
Adds three different implementations for attentions that can be used for text classification

- Self attention is the one used in BERT
- Hierarchical attention is the one used in Spacy
- Feed forward attention is a simplest form of the above and quite a common depiction of attention

I am adding hierarchical attention to the models with a param that defaults to False so someone can easily enable and experiment whether it improves results. The reason it does not default to True is that I am getting some mixed results in the HOC dataset, possibly due to different hyper parameters that are needed. Will keep exploring this

### DIFF
--- a/wellcomeml/ml/attention.py
+++ b/wellcomeml/ml/attention.py
@@ -1,4 +1,7 @@
+import tensorflow as tf
+
 class SelfAttention(tf.keras.layers.Layer):
+    """https://papers.nips.cc/paper/7181-attention-is-all-you-need.pdf"""
     def __init__(self, attention_dim=20):
         super(SelfAttention, self).__init__()
         self.attention_dim = attention_dim
@@ -9,9 +12,10 @@ class SelfAttention(tf.keras.layers.Layer):
         self.WV = self.add_weight(shape=(input_shape[-1], input_shape[-1]), trainable=True, initializer='uniform')
 
     def call(self, X):
-    	"""
-    	X: (batch_size, sequence_length, embedding_dimension)
-    	"""
+        """
+        In: (batch_size, sequence_length, embedding_dimension)
+        Out: (batch_size, sequence_length, embedding_dimension)
+        """
         Q = tf.matmul(X, self.WQ)
         K = tf.matmul(X, self.WK)
         V = tf.matmul(X, self.WV)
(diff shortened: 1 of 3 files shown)


nsorros:
Should have opened as draft. The problem with attention is that it does not seem to improve results as it is. So need to investigate what might be missing.

nsorros:
@lizgzil @aCampello ready for review

wellcomeml/ml/attention.py
```diff
@@ -0,0 +1,57 @@
+import tensorflow as tf
+
+class SelfAttention(tf.keras.layers.Layer):
+    """https://papers.nips.cc/paper/7181-attention-is-all-you-need.pdf"""
+    def __init__(self, attention_dim=20):
+        super(SelfAttention, self).__init__()
```
lizgzil, line 6: super seems handy!

wellcomeml/ml/cnn.py
```diff
@@ -49,6 +52,12 @@ def residual_conv_block(x1):
             x2 = tf.keras.layers.LayerNormalization()(x2)
             return tf.keras.layers.add([x1, x2])
 
+        def residual_attention(x1):
+            x2 = HierarchicalAttention()(x1)
```
lizgzil, line 56: How come you initialise this but you don't do `build` or `call`?

nsorros: It happens under the hood from tensorflow

nsorros: https://www.tensorflow.org/guide/keras/custom_layers_and_models

lizgzil: oo nice

nsorros:
> I really like the fact that we have our own CNN classifier and can add all theses layers. One thing I wasn't unsure is how this custom layer relates to keras' attention layer (https://www.tensorflow.org/api_docs/python/tf/keras/layers/Attention). Is it because of self-attention ?

I hope I answered during the call but will repeat in case it was not clear and to have a log. Keras attention layer assumes you have two vectors Q, V and optionally K. The two vectors can come from the output of the encoder and decoder of bilstm in which case it makes sense to use this layer. In the case of text classification we only have one vector the vector that corresponds to a token for each time step. The other vector is created using some weights so this process of creating the new vector and doing the dot product etc is incorporated into this attention implementation. One way I could have used the attention layer is to create the vector and then pass the two vectors from the attention layer inside the implementation. But doing the dot product and softmax is not the complicated bit.

nsorros:
[Add attention that can be used in both CNN and biLSTM architectures](https://trello.com/c/BtWzNJgr/964-add-attention-that-can-be-used-in-both-cnn-and-bilstm-architectures)

PR #41, opened 2020-04-22
Add hoc dataset

nsorros:
Adds HOC dataset https://github.com/sb895/Hallmarks-of-Cancer.

Introduces a dataset folder that aims to standardise downloading standard datasets that we want to experiment on.

### DIFF
--- a/wellcomeml/datasets/download.py
+++ b/wellcomeml/datasets/download.py
@@ -1,0 +1,45 @@
+import os
+import tarfile
+
+import boto3
+from botocore import UNSIGNED
+from botocore.client import Config
+
+from wellcomeml.logger import logger
+
+DATA_DIR = os.path.expanduser("~/.cache/wellcomeml/data")
+DATA_DISPATCH = {
+    "hoc": {
+        "bucket": "datalabs-public",
+        "path": "datasets/hoc/hoc.tar",
+        "file_name": "hoc.tar"
+    }
+}
+
+def check_cache_and_download(dataset_name):
+    """ Checks if dataset_name is cached and return complete path"""
+    os.makedirs(DATA_DIR, exist_ok=True)
+
+    dataset_path = os.path.join(DATA_DIR, dataset_name)
+    if not os.path.exists(dataset_path):
+        logger.info(f"Could not find dataset {dataset_name}. Downloading from S3")
+
+        # The following allows to download from S3 without AWS credentials
+        s3 = boto3.client('s3', config=Config(signature_version=UNSIGNED))
+        tmp_file = os.path.join(
+            DATA_DIR,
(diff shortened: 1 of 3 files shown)


wellcomeml/datasets/download.py
```diff
@@ -0,0 +1,45 @@
...
+DATA_DIR = os.path.expanduser("~/.cache/wellcomeml/data")
+DATA_DISPATCH = {
+    "hoc": {
+        "bucket": "datalabs-public",
+        "path": "datasets/hoc/hoc.tar",
+        "file_name": "hoc.tar"
+    }
+}
+
+def check_cache_and_download(dataset_name):
```
lizgzil, line 19: handy function!

lizgzil: **thought**: I wonder whether it'd be good to keep it really generic so that we could use it for other S3 downloads too? e.g. have arguments for DATA_DIR and DATA_DISPATCH  and allow the case where credentials are needed?

wellcomeml/datasets/hoc.py
```diff
@@ -0,0 +1,41 @@
...
+
+    train_data_path = os.path.join(path, "train.tsv")
+    X_train, Y_train = load_split(train_data_path)
+
+    test_data_path = os.path.join(path, "test.tsv")
+    X_test, Y_test = load_split(test_data_path)
+
+    if split == 'train':
+        X = X_train
+        Y = Y_train
```
lizgzil, line 28: **suggestion**: I guess you are unnecessarily loading and splitting the test data in this case (and vice versa for split='test') consider only loading when it's needed

aCampello: +1 for that suggestion.

wellcomeml/datasets/download.py
aCampello: Note that https://github.com/wellcometrust/WellcomeML/blob/master/wellcomeml/ml/bert_vectorizer.py#L99-L121, we already use it for downloading SciBERT models too, so we might consider generalising it in a later work.

wellcomeml/datasets/hoc.py
nsorros: it was done to avoid repetition of code but will revert

aCampello:
Is there anything else to merge this PR?

nsorros:
> Is there anything else to merge this PR?

Yes, implementing the change that Liz suggested. Will do that now :P

nsorros:
[Add dataset for biomedical text classification and create a benchmark that gets results on the dataset](https://trello.com/c/qyWcZubf/966-add-dataset-for-biomedical-text-classification-and-create-a-benchmark-that-gets-results-on-the-dataset)

PR #43, opened 2020-04-24
Semantic similarity

aCampello:
This adds a base code to train a document similarity model. It includes

- [x] Similarity model for transformers/tensorflow (`ml/bert_semantic_equivalence.py`)
- [x] Tests and independent test platform for new versions of transformers that clash with spacy
- [x] An example with 100 pairs of sentences

To test:
```bash
make test-transformers
```


aCampello:
[Implement a Sentence Matching tuning model for Scibert classifier](https://trello.com/c/qob2Tn2T/927-implement-a-sentence-matching-tuning-model-for-scibert-classifier)

tests/test_semantic_similarity.py
```diff
@@ -0,0 +1,27 @@
...
+         ('This sentence is about something else', 'God save the queen')]
+
+    y = [1, 1, 0]
+
+    classifier.fit(X, y, epochs=3)
+
+    losses = classifier.model.history.history['loss']
+    scores = classifier.score(X)
+
+    assert losses[2] < losses[0]
```
nsorros, line 24: 👍 maybe we should generally move towards comparing loss in all models

wellcomeml/ml/bert_semantic_equivalence.py
```diff
@@ -0,0 +1,176 @@
...
+import tensorflow as tf
+
+from transformers import (
+    BertConfig,
+    BertTokenizer,
+    TFBertForSequenceClassification,
+    glue_convert_examples_to_features,
+    glue_processors,
+)
+
```
nsorros, line 12: why the spaces in between the external libraries?

wellcomeml/ml/bert_semantic_equivalence.py
```diff
@@ -0,0 +1,176 @@
...
+    # Bert models have a tensorflow checkopoint, otherwise,
+    # we need to load the pytorch versions with the parameter `from_pt=True`
+
+    def _initialise_models(self):
+        if self.pretrained == 'bert':
+            model_name = 'bert-base-cased'
+            from_pt = False
+        elif self.pretrained == 'scibert':
+            model_name = check_cache_and_download('scibert_scivocab_cased')
```
nsorros, line 54: why do we need to use this since scibert is part of transformers?

wellcomeml/ml/bert_semantic_equivalence.py
```diff
@@ -0,0 +1,176 @@
...
+            from_pt = True
+
+        print(model_name)
```
nsorros, line 57: logger maybe better?

wellcomeml/ml/bert_semantic_equivalence.py
nsorros: is it because the one that comes from them is pytorch?

wellcomeml/ml/bert_semantic_equivalence.py
```diff
@@ -0,0 +1,176 @@
...
+        # Train and evaluate using tf.keras.Model.fit()
+
+        metric = tf.keras.metrics.SparseCategoricalAccuracy("accuracy")
+
+        self.model.compile(optimizer=opt, loss=loss, metrics=[metric])
+
+        self.train_steps = len(X_train) // self.batch_size
+        self.valid_steps = len(X_valid) // self.eval_batch_size
+
+        self.fit_epoch(epochs=epochs)
```
nsorros, line 137: why the need for a separate function?

wellcomeml/ml/bert_semantic_equivalence.py
```diff
@@ -0,0 +1,176 @@
...
+            validation_data=self.valid_dataset,
+            validation_steps=self.valid_steps,
+        )
+
+        return self
+
+    def score(self, X):
+        X_tokenized = self.tokenizer.batch_encode_plus(
+            X, max_length=self.max_length, add_special_tokens=True,
+            pad_to_max_length=True,
```
nsorros, line 159: what if the X passed is of different length to the train one?

wellcomeml/ml/bert_semantic_equivalence.py
```diff
@@ -0,0 +1,176 @@
...
+            return_tensors="tf"
+        )
+        predictions = tf.convert_to_tensor(self.model.predict(X_tokenized))
```
nsorros, line 162: predict should probably operate on text as well.

lizgzil:
I''m getting the error
```
ERROR: spacy-transformers 0.5.1 has requirement transformers<2.1.0,>=2.0.0, but you'll have transformers 2.4.1 which is incompatible.
Installing collected packages: filelock, tokenizers, transformers, tensorflow
  Attempting uninstall: transformers
    Found existing installation: transformers 2.0.0
    Uninstalling transformers-2.0.0:
      Successfully uninstalled transformers-2.0.0
Successfully installed filelock-3.0.12 tensorflow-2.1.0 tokenizers-0.0.11 transformers-2.4.1
build/virtualenv/bin/pytest -m "transformers" --disable-warnings --cov-append --tb=line --cov=wellcomeml ./tests
/bin/sh: build/virtualenv/bin/pytest: No such file or directory
```
when I do `make test-transformers`

wellcomeml/ml/bert_semantic_equivalence.py
```diff
@@ -0,0 +1,176 @@
...
+
+        return tf.keras.activations.softmax(predictions).numpy()
+
+    def predict(self, X):
```
nsorros, line 166: X should be the same as fit in my opinion

wellcomeml/ml/bert_semantic_equivalence.py
```diff
@@ -0,0 +1,176 @@
...
+        # Initialises/downloads model
+        self._initialise_models()
+
+        # Train/val split
+        X_train, X_valid, y_train, y_valid = train_test_split(
+            X, y, test_size=self.test_size, random_state=random_state
+        )
+
+        # Generates tensorflow dataset
+        self.train_dataset = self._prep_dataset_generator(X_train, y_train)
```
nsorros, line 115: generators for the win

wellcomeml/ml/bert_semantic_equivalence.py
```diff
@@ -0,0 +1,176 @@
...
+
+
+class SemanticEquivalenceClassifier(BaseEstimator, TransformerMixin):
+    """
+    Class to fine-tune BERT-type models for semantic equivalence, for example
+    paraphrase, textual similarity and other NLU tasks
+    """
+    def __init__(self, pretrained='bert',
+                 batch_size=32,
+                 eval_batch_size=32*2,
```
nsorros, line 26: why different?

wellcomeml/ml/bert_semantic_equivalence.py
aCampello: Yes - because at the moment the scibert model on the hub unfortunately cannot be fine-tuned with `TFBertForSequenceClassification`, because it only contains the pytorch weights, and you cannot load `from_pt` with pre_trained models in the hub.

aCampello:
> I''m getting the error
> 
> ```
> ERROR: spacy-transformers 0.5.1 has requirement transformers<2.1.0,>=2.0.0, but you'll have transformers 2.4.1 which is incompatible.
> Installing collected packages: filelock, tokenizers, transformers, tensorflow
>   Attempting uninstall: transformers
>     Found existing installation: transformers 2.0.0
>     Uninstalling transformers-2.0.0:
>       Successfully uninstalled transformers-2.0.0
> Successfully installed filelock-3.0.12 tensorflow-2.1.0 tokenizers-0.0.11 transformers-2.4.1
> build/virtualenv/bin/pytest -m "transformers" --disable-warnings --cov-append --tb=line --cov=wellcomeml ./tests
> /bin/sh: build/virtualenv/bin/pytest: No such file or directory
> ```
> 
> when I do `make test-transformers`

I think you need to create the virtualenv witth `make virtualenv` first. You might have to re-created if you created it a while ago, because it installs the pytest routines.

wellcomeml/ml/bert_semantic_equivalence.py
aCampello: Yes this slipped through the code

wellcomeml/ml/bert_semantic_equivalence.py
aCampello: I wanted to have a function that only fits an extra epoch, without having to pass in any data or anything, so I can skip all transformations. I think this could actually be achieved in one function - will look into it 👍

wellcomeml/ml/bert_semantic_equivalence.py
aCampello: the class predict does operate on text. So for `self.predict` you pass a list o 2-uples of texts and it will output whether the text is related (0) or not (1). The tensorflow model (which the end-user will probably not have to use) receives tokenised texts.

I added a docstring in https://github.com/wellcometrust/WellcomeML/pull/43/commits/eb5b9d25d6962ae44ecb52e51f6114a089cca10c for clarification.

wellcomeml/ml/bert_semantic_equivalence.py
aCampello: Indeed, it is - sorry if it wasn't clear before. Please see the new docstrings (https://github.com/wellcometrust/WellcomeML/pull/43/commits/eb5b9d25d6962ae44ecb52e51f6114a089cca10c)

wellcomeml/ml/bert_semantic_equivalence.py
aCampello: That is ok - if it is bigger than max_length it will parse it, and if it is smaller, it will pad it. Note that `self.max_length` is "global" to the class, and only defined once at training time.

wellcomeml/ml/bert_semantic_equivalence.py
aCampello: Good question - I don't think these too are tied, and for small sets `eval_batch_size` can even be set to the entire dataset. I suppose `eval_batch_size` can be bigger because it doesn't to backpropagation (just a forward pass). For different datasets it will be a question of experimentation I guess.

wellcomeml/ml/bert_semantic_equivalence.py
nsorros: Is that an issue that can be opened in hugging face repo?

wellcomeml/ml/bert_semantic_equivalence.py
aCampello: This was fixed in the new commits.

wellcomeml/ml/bert_semantic_equivalence.py
aCampello: Fixed in https://github.com/wellcometrust/WellcomeML/pull/43/commits/cca45683f07aedc492664e29db13477ab91284eb, https://github.com/wellcometrust/WellcomeML/pull/43/commits/ef399baaf9db8c67c20c441e35f6b74cf88c0f9f

wellcomeml/ml/bert_semantic_equivalence.py
aCampello: This has been fixed in https://github.com/wellcometrust/WellcomeML/pull/43/commits/3d770ba1d5232ec33f3fcf3f2acf7b88bb4536d6

wellcomeml/ml/bert_semantic_equivalence.py
aCampello: Turns out this is was solved in transformers `2.6.0`. I bumped the version up and was able to get rid of check_cache in https://github.com/wellcometrust/WellcomeML/pull/43/commits/fd56cbed3cf0d036c304ee51f097695a6e8c4a6e 🎉

aCampello:
Would you mind having another look @lizgzil @nsorros ? I think I implemented all the suggestions.

wellcomeml/ml/bert_semantic_equivalence.py
```diff
@@ -52,7 +52,7 @@ def _initialise_models(self):
             model_name = 'bert-base-cased'
             from_pt = False
         elif self.pretrained == 'scibert':
-            model_name = check_cache_and_download('scibert_scivocab_cased')
+            model_name = model_name = 'allenai/scibert_scivocab_cased'
```
nsorros: is this line `model_name = model_name = 'allenai/scibert_scivocab_cased'`?

wellcomeml/ml/bert_semantic_equivalence.py
nsorros: Interesting, i have not seen the differentiation before. I am not so sure this parameter changes anything, maybe it makes inference faster in case data fit into memory.

wellcomeml/ml/bert_semantic_equivalence.py
nsorros: just got confused with the two predicts. it makes sense.

wellcomeml/ml/bert_semantic_equivalence.py
```diff
@@ -0,0 +1,198 @@
...
+        self.config = None
+        self.tokenizer = None
+        self.model = None
+        self.train_dataset = None
+        self.valid_dataset = None
+        self.train_steps = None
+        self.valid_steps = None
+        
+        # Flag to determine whether a model was previously trained or not
+        self._trained = False
```
nsorros, line 44: possibly it would be nice if this was sklearn compatible `https://scikit-learn.org/stable/modules/generated/sklearn.utils.validation.check_is_fitted.html`, i think it would need an undescore in the end.

wellcomeml/ml/bert_semantic_equivalence.py
```diff
@@ -0,0 +1,198 @@
...
+
+    def score(self, X):
+        """
+        Calculates scores for model prediction
+
+        Args:
+            X: List of 2-uples: [('Sentence 1', 'Sentence 2'), ....]
+
+        Returns:
+            An array of shape len(X) x 2 with scores for classes 0 and 1
```
nsorros, line 165: doesn't this return an array of len(X)?

aCampello:
> 
> I also got the error `ERROR: spacy-transformers 0.5.1 has requirement transformers<2.1.0,>=2.0.0, but you'll have transformers 2.6.0 which is incompatible.` when I was building the transformers in . Which I guess you know about!

Yes, the error is intentionally induced at this point, to switch spacy off. I'll look into adding something about the rust error to readme

wellcomeml/ml/bert_semantic_equivalence.py
aCampello: Yeah - I fixed that in a subsequent commit https://github.com/wellcometrust/WellcomeML/pull/43/commits/5f38110846cc2574535089921a976eff26652130

wellcomeml/ml/bert_semantic_equivalence.py
aCampello: That is a really good suggestion - I will look into it before merging.

wellcomeml/ml/bert_semantic_equivalence.py
aCampello: No, this returns two scores for every data point (probably of 0 and probability of 1). Possibly redundant (could do with one score), but it is the usual softmax output for Keras.

wellcomeml/ml/bert_semantic_equivalence.py
aCampello: I implemented this in https://github.com/wellcometrust/WellcomeML/pull/43/commits/750e13a46fde40a2b3c94f3334e837b9d7537b2f

PR #44, opened 2020-05-04
Improve entity linking

lizgzil:
Updating the entity linking code with the new and improved TFIDF/BERT based entity linking. The old version used functions from spacy and had stopped being able to be run due to needing an older version of spacy than is currently in wellcomeml.

The code in `entity_linking.py` was previously in the `datalabs` repo [here](https://github.com/wellcometrust/datalabs/blob/master/modelling/policy/entity-finder/entity_finder/tfidf_entity_linker.py), but I am moving it over to live here. See PR https://github.com/wellcometrust/datalabs/pull/684/files for the work of removing it from datalabs. The only difference is that the class was called `TfidfLinker` in datalabs but here it is called `EntityLinker`, which is more suitable anyway since it has the ability to use BERT embeddings instead of TFIDF.

Note:
Because `EntityLinker` imports `BertVectorizer` the order of importing these 2 in the init file is important (BertVectorizer before EntityLinker).


nsorros:
A small limitation of the current approach is that it does not learn much, i.e. it basically fits TFIDF or uses BERT to embed. This is not necessarily a problem but in the future you can look into what Antonio built for reranking, basically given some training data you can rerank the retrieved entities using a neural model or other model.

nsorros:
Another comment is that you should consider whether you want to treat this problem as a 2 step problem as it was before. Step one fitting a knowledge base, in which the embedding would happen, either from scratch with TFIDF or some a neural embedding layer or with some pre trainined embeddings like frozen bert. Step two fitting an entity linker, in which you get a text with entities and correct entities and then you train to differentiate which entity is from the knowledge base (also an input, possibly in the init). At the moment you are doing both steps in one. This might get slighly confusing as your entity linker behaves a bit more like a knowledge base, i.e. you need to pass information about entities into fit in order to get predictions in predict. But it could be that you believe that this is more intuitive and on purpose. In order to decide you might want to think what happens in a future point where you get annotated entities with links, how do you use them? currently you need to find information on the entities and pass them to fit and then use the annotations to evaluate (but not train).

examples/entity_linking.py
```diff
@@ -0,0 +1,33 @@
+from wellcomeml.ml import EntityLinker
+
+entities_kb = {
+        'Michelle Williams (actor)': "American actress. She is the recipient of several accolades, including two Golden Globe Awards and a Primetime Emmy Award, in addition to nominations for four Academy Awards and one Tony Award.",
+        'Michelle Williams (musician)': "American entertainer. She rose to fame in the 2000s as a member of R&B girl group Destiny's Child, one of the best-selling female groups of all time with over 60 million records, of which more than 35 million copies sold with the trio lineup with Williams.",
```
nsorros, line 5: this representation might be a bit limited. how will you construct a unique id? will you keep the id separately? i.e. Jeremmy Farrar might not be unique, if you replace with ORCID id where will you keep the association between ORCID and name? it wold be nice if the knowledge base contained this information, essentially the value was not text but a dict that contained text along with other information

lizgzil: The `entities_kb` dictionary key is the unique id, and can be anything as long as it is consistent with what is used in the tag (e.g. `{'id': 'Michelle Williams (musician)'}`), so this could be the ORCID or another ID you want to use. I just put their names so the print out would be clearer.

In terms of using ORCIDs I wouldn't have 'Jeremy Farrar', I'd have his ORCID.

If you think this point isn't clear in this example I can change it to:
```
entities_kb = {
        'id_1': "American actress. She ....",
        'id_2': "American entertainer. She....",
        'id_3': "  "
        }
train_data = [
        ("After Destiny's ...", {'id': 'id_2'}),
        ("On Broadway ....", {'id': 'id_1'}),
        ("Franklin would ...", {'id': 'No ID'})
        ]
```

lizgzil:
@nsorros I do agree that it might be good to do the knowledge base curation separately, it does feel like this class is trying to do 2 separate things.
However, the purpose of this PR was really just to move the code from `datalabs` to `wellcomeml` (on my to do list for ages!), I don't intend to work on entity linking for a while. So if the code is ok I would prefer to get it merged as is, and then when I readdress the entity linking (which could be in a few months) I will probably want to restructure things then anyway.

nsorros:
> @nsorros I do agree that it might be good to do the knowledge base curation separately, it does feel like this class is trying to do 2 separate things.
> However, the purpose of this PR was really just to move the code from `datalabs` to `wellcomeml` (on my to do list for ages!), I don't intend to work on entity linking for a while. So if the code is ok I would prefer to get it merged as is, and then when I readdress the entity linking (which could be in a few months) I will probably want to restructure things then anyway.

Fair enough, do you think it might be worth keeping the code for SpacyEntintyLinking and SpacyKnowledgeBase and name this somehow else? For example SimilarityEntityLinking

lizgzil:
@nsorros yep good idea, i've made the name changes and added back in the spacy scripts

PR #50, opened 2020-05-13
WellcomeML to PyPi

nsorros:
This is a WIP PR to move WellcomeML to PyPi.

We need to decide on two things:

- which email to use to create an account in pypi
- which method to use to download models that needed to be removed from dependencies in order to upload to pypi

One way we can download models is through a command in Makefile which I have implemented. This is not different experience from the one you have at spacy where you need to install the library and download models separately. We could implement a helper download function like in spacy and we can incorporate download when a model is needed into the codebase as transformers do and as we do for scispacy. I am happy with either, I favour the last.


aCampello:
Maybe `Grp_datalabs-datascience@Wellcomecloud.onmicrosoft.com` for the account? 

I would prefer to have a built-in function to download the model. This gives us flexibility and means the users don't need to rely on a makefile recipe.

nsorros:
> Maybe `Grp_datalabs-datascience@Wellcomecloud.onmicrosoft.com` for the account?
> 
> I would prefer to have a built-in function to download the model. This gives us flexibility and means the users don't need to rely on a makefile recipe.

In regards to the two built in download options, which one do we favour? Something like spacy where the user explicitly downloads the models or something like huggingface where a check is performed and if the model is missing it gets downloaded?

aCampello:
Seems that `make test` failed: 

```
$ make test
558build/virtualenv/bin/python -m wellcomeml download models
559/home/travis/virtualenv/python3.7.1/bin/python: No module named spacy
560touch build/virtualenv/.models
561make: *** No rule to make target 'build/virtualenv/.deeplearning-models', needed by 'test'.  Stop.
562The command "make test" exited with 2.
```

lizgzil:
> > Maybe `Grp_datalabs-datascience@Wellcomecloud.onmicrosoft.com` for the account?
> > I would prefer to have a built-in function to download the model. This gives us flexibility and means the users don't need to rely on a makefile recipe.
> 
> In regards to the two built in download options, which one do we favour? Something like spacy where the user explicitly downloads the models or something like huggingface where a check is performed and if the model is missing it gets downloaded?

I don't have strong feelings about this, but my input is that the check+download option is a little less transparent than the user having to explicitly state which model to download (i.e. spacy).

nsorros:
> > > Maybe `Grp_datalabs-datascience@Wellcomecloud.onmicrosoft.com` for the account?
> > > I would prefer to have a built-in function to download the model. This gives us flexibility and means the users don't need to rely on a makefile recipe.
> > 
> > 
> > In regards to the two built in download options, which one do we favour? Something like spacy where the user explicitly downloads the models or something like huggingface where a check is performed and if the model is missing it gets downloaded?
> 
> I don't have strong feelings about this, but my input is that the check+download option is a little less transparent than the user having to explicitly state which model to download (i.e. spacy).

is less transparent good or bad?

lizgzil:
> > > > Maybe `Grp_datalabs-datascience@Wellcomecloud.onmicrosoft.com` for the account?
> > > > I would prefer to have a built-in function to download the model. This gives us flexibility and means the users don't need to rely on a makefile recipe.
> > > 
> > > 
> > > In regards to the two built in download options, which one do we favour? Something like spacy where the user explicitly downloads the models or something like huggingface where a check is performed and if the model is missing it gets downloaded?
> > 
> > 
> > I don't have strong feelings about this, but my input is that the check+download option is a little less transparent than the user having to explicitly state which model to download (i.e. spacy).
> 
> is less transparent good or bad?

I think it's bad

aCampello:
This is huge, thanks a lot Nick. I am happy to approve the PR, just two extra points:

- [ ] Upload to pipit
- [ ] Can we update README.md 

I am happy to help with that if you like.

wellcomeml/__main__.py
```diff
@@ -0,0 +1,26 @@
+import subprocess
+import sys
+import os
+
+
+VENV_PYTHON = os.path.join(
```
aCampello, line 6: I'd still like `VENV_PYTHON` to be able to be configured by the user (and possibly defaulted to plain `python`). The problem with the current approach is that users will have to have a virtualenv named exactly `build/virtualenv/bin/python`. For example. if I install wellcomeml in a virtualenv called `venv`, this command will not allow me to download models...

aCampello:
FYI @ivyleavedtoadflax

ivyleavedtoadflax:
Great work everyone 👍👍👍

wellcomeml/__main__.py
nsorros: hmm, that is a very good point. in this case i think we should go back to python and delegate to the user i.e. if the user is in a virtualenv then it will install there otherwise it will install wherever python is. that makes more sense. the make command should assume venv is created with make so that is fine as well. the only problem to resolve are the tests in travis.

nsorros:
> This is huge, thanks a lot Nick! I am happy to approve the PR, just two extra points:
> 
> * [ ]  Upload to pypi
> * [ ]   Can we update README.md we new pypi installation instructions
> 
> I am happy to help with that if you like, or to defer these steps to a subsequent PR. Also have a look at the comment below about the VENV_PYTHON variable being configurable.

Will do on Monday and upload

wellcomeml/__main__.py
nsorros: one way to fix that is to change the make command actually to not use the download utility and install models which would be duplicate and we would need to maintain which models are downloaded in two places. alternatively i am wondering whether we can prepend the activation of the environment as a hack?

aCampello: That's true. I think it is ok to assume the user will be in the correct environment if he runs the command ad-hoc. As for the makefile, we can replace 

https://github.com/wellcometrust/WellcomeML/blob/b52fb5cd1864ff04eff4488ae432727cd84c0996/Makefile#L44

to
```bash
. $(VIRTUALENV)/bin/python && $(VIRTUALENV)/bin/python -m spacy download en_core_web_sm
```
I believe it should work, but it has to be done in the same line, because otherwise `make` will run it in a different subprocess.

aCampello:
:shipit:

PR #55, opened 2020-05-25
Add Doc2VecVectorizer

nsorros:
Add Doc2Vec vectorizer by wrapping up gensim into a sklearn compatible class.


nsorros:
[Add Doc2VecVectorizer in WellcomeML](https://trello.com/c/2yUTweZ8/992-add-doc2vecvectorizer-in-wellcomeml)

tests/test_doc2vec.py
```diff
@@ -0,0 +1,30 @@
+from wellcomeml.ml import Doc2VecVectorizer
+
+def test_fit_transform():
+    X = [
+        "Wellcome trust gives grants",
+        "Covid is a infeactious disease",
+        "Sourdough bread is delicious",
```
lizgzil, line 7: 👏

examples/doc2vec.py
```diff
@@ -0,0 +1,25 @@
...
+
+X = [
+    "Malaria is a disease spread by mosquitos",
+    "HIV is a virus that causes a disease named AIDS",
+    "Trump is the president of USA"
+]
+
+doc2vec = Doc2VecVectorizer(min_count=1, vector_size=8)
+X_transformed = doc2vec.fit_transform(X)
+print(cosine_similarity(X_transformed))
```
lizgzil, line 16: I get this result for this 
```
[[ 1.0000001  -0.08753361  0.03297466]
 [-0.08753361  1.          0.1474569 ]
 [ 0.03297466  0.1474569   0.9999999 ]]
```
I would have thought sentences 0 and 1 would be quite similar, but this is suggesting that sentences 1 and 2 are more similar than 0 and 1.

I has a play with changing X to 
```
X = [
    "Trump is the president of US",
    "Trump is the president of USA",
    "Malaria is a disease spread by mosquitos",
]
```
and got 
```
[[ 1.          0.2648901  -0.01699547]
 [ 0.2648901   1.         -0.41694677]
 [-0.01699547 -0.41694677  1.        ]]
```
which seems a bit odd? why is the cosine so small (0.26)?

nsorros: You are 100% right, i did not manage to find a good example yet. Might give it another go before merging but the main problem seems to be that the model does not produce meaningful embeddings for such trivial data. There are similar comments made by others online. That said it would be nice to find an illustrative example that works

wellcomeml/ml/doc2vec_vectorizer.py
```diff
@@ -0,0 +1,127 @@
...
+        return statistics.mean(correct)
+
+    def save(self, model_path):
+        self.model.save(model_path)
+
+    def load(self, model_path):
+        self.model = Doc2Vec.load(model_path)
+
+if __name__ == '__main__':
+    DATA = "data.jsonl"
```
nsorros, line 112: oups this needs to go before merging

examples/doc2vec.py
nsorros: I improved the example slighly in that the first two sentences are consistently closer than the third. I have not found a way to fix your example, the best case gives me a 0.5 similarity. This is weird. I opened an issue https://github.com/wellcometrust/WellcomeML/issues/60

PR #56, opened 2020-06-02
Add NER datasets

lizgzil:
Trello card: https://trello.com/c/4f6TI0uP/996-adding-ner-dataset-to-wellcomeml

Note: `winer.py` and `conll.py` are intended to follow a similar layout to `hoc.py`.

I have added `s3://datalabs-public/datasets/ner/winer.tar` and `s3://datalabs-public/datasets/ner/conll.tar` NER datasets (from a [wikipedia corpus](https://www.aclweb.org/anthology/I17-1042/) and a [Reuters corpus](https://www.clips.uantwerpen.be/conll2003/ner/) respectively).

`conll.py` loads the CONLL dataset.

`winer.py` loads the WiNER dataset. This also has the function `create_train_test` to process the WiNER raw data which comes in 3 files that need to be merged together to create train.txt and test.txt files (not included in the raw data). I've run this already and uploaded the train and test txt files to the winer.tar file, so you shouldnt need to run the `create_train_test` function again, but it is included for transparency.

## How to test the datasets
You can now run `python examples/train_ner_spacy.py` to get examples of the spacy ner model being trained on custom (defined in the script), the conll and the winer dataset.

## Next
- Use these new datasets to train in datalabs with the tagged policy data as the evaluation data.


wellcomeml/datasets/winer.py
```diff
@@ -0,0 +1,222 @@
...
+
+    print("Creating entities dictionary from {}".format(NE_path))
+    entities = {}
+    with tarfile.open(NE_path, "r:bz2") as tar:
+        for member in tqdm(tar.getmembers()):
+            f = tar.extractfile(member)
+            if f:
+                content = f.read().decode('utf-8', errors='ignore')
+                articles = content.replace('\t', ' ').split('ID ')
+                if articles[0]=='':
```
aCampello, line 56: Is that a problem with their file or with the way we read the file?

wellcomeml/datasets/winer.py
```diff
@@ -0,0 +1,222 @@
...
+                    entities[article_id] = sentence_entities
+
+    # Create id2word from vocab file (line number is id)
+    print("Creating id2word dictionary from {}".format(vocab_path))
+    id2word = {}
+    with open(vocab_path, "r") as vocab:
+        f = vocab.read()
+        lines = f.split('\n')
+        for i, line in enumerate(lines):
+            id2word[i] = line.split(' ')[0]
```
aCampello, line 82: A minor remark: This dictionary could be simply a list, since you are just attributing numbers from `0` to `len(lines)`. Then the entries could be accessed the same way by `id2word[i]`

wellcomeml/datasets/winer.py
```diff
@@ -0,0 +1,222 @@
...
+
+    ID 444
+    Next
+    document
+    """
+
+    random.seed(rand_seed)
+    n_train = round(prop_train*n_sample)
+
+    print("Creating entities dictionary from {}".format(NE_path))
```
aCampello, line 48: Can we do logging? `from wellcomeml.logger import logger`

wellcomeml/datasets/winer.py
```diff
@@ -0,0 +1,222 @@
...
+                                tokens = [id2word[int(s)] for s in sent_wordidx[int(sentid)]]
+                                tags = ['O']*len(tokens)
+                                for begin, end, ent_type in sentence_entities:
+                                    tags[(begin+1):(end-1)] = [str(ent_type) + '-I']*(end-1-begin-1)
+                                    tags[end-1] = str(ent_type) + '-E'
+                                    tags[begin] = str(ent_type) + '-B'
+                                for token, tag in list(zip(tokens, tags)):
+                                    output_file.write(' '.join([token, tag]))
+                                    output_file.write('\n')
+                        output_file.write('\n')
```
aCampello, line 123: I am not 100% sure of the logic, but it seems that is is outputting a double `\n` for the last entity?

wellcomeml/datasets/conll.py
```diff
@@ -0,0 +1,56 @@
+from wellcomeml.datasets.download import check_cache_and_download
+
+import os
+import random
+
+def load_data_spacy(data_path, inc_outside=True):
```
nsorros, line 6: maybe you want this to start with an underscore

wellcomeml/datasets/conll.py
```diff
@@ -0,0 +1,56 @@
...
+        for line in lines:
+            article_text = ''
+            char_i = 0
+            article_tags = []
+            entities = line.split('\n')
+            for entity in entities:
+                if len(entity) != 0:
+                    token, _, _, tag = entity.split(' ')
+                    article_text += token + ' '
+                    if tag!='O' or inc_outside:
```
nsorros, line 26: `tag != 0` space missing

wellcomeml/datasets/conll.py
```diff
@@ -0,0 +1,56 @@
...
+    # Load data in Spacy format:
+    # X = list of sentences (plural) / documents ['the cat ...', 'some dog...', ...]
+    # Y = list of list of entity tags for each sentence [[{'start': 36, 'end': 46, 'label': 'PERSON'}, {..}, ..], ... ]
+    # inc_outside = False: don't include none-entities in the output
+
+    X = []
+    Y = []
+    with open(data_path) as f:
+        lines = f.read().split('-DOCSTART- -X- O O\n')
+        for line in lines:
```
nsorros, line 17: i think these are not lines, the are possibly articles, i suggest you name them accordingly

wellcomeml/datasets/conll.py
```diff
@@ -0,0 +1,56 @@
...
+            article_tags = []
+            entities = line.split('\n')
+            for entity in entities:
+                if len(entity) != 0:
+                    token, _, _, tag = entity.split(' ')
+                    article_text += token + ' '
+                    if tag!='O' or inc_outside:
+                        article_tags.append({'start': char_i, 'end': char_i+len(token), 'label': tag})
+                    char_i += len(token) + 1 # plus 1 for the space separating
+            if article_tags!=[]:
```
nsorros, line 29: space missing

wellcomeml/datasets/winer.py
```diff
@@ -0,0 +1,222 @@
...
+from wellcomeml.datasets.download import check_cache_and_download
+
+# def load_winer():
+import tarfile
+import os
+from tqdm import tqdm 
+import json
+import random
+
+def create_train_test(
```
nsorros, line 13: is this the function that takes the source data and transform them to train and test? could that be mentioned more explicitly in the comment section below as this is a function that someone might need to use, if say we want to recreate the files since you do not store them in the raw format.

wellcomeml/datasets/winer.py
```diff
@@ -0,0 +1,222 @@
...
+                                for begin, end, ent_type in sentence_entities:
+                                    tags[(begin+1):(end-1)] = [str(ent_type) + '-I']*(end-1-begin-1)
+                                    tags[end-1] = str(ent_type) + '-E'
+                                    tags[begin] = str(ent_type) + '-B'
+                                for token, tag in list(zip(tokens, tags)):
+                                    output_file.write(' '.join([token, tag]))
+                                    output_file.write('\n')
+                        output_file.write('\n')
+
+def load_data_spacy(data_path, inc_outside=True, merge_entities=True):
```
nsorros, line 125: maybe an underscore

wellcomeml/datasets/winer.py
```diff
@@ -0,0 +1,222 @@
...
+    # Y = list of list of entity tags for each sentence [[{'start': 36, 'end': 46, 'label': 'PERSON'}, {..}, ..], ... ]
+    # inc_outside = False: don't include none-entities in the output
+    # merge_entities if entities span over multiple tags do you want to merge them or not
+    # e.g. John 3-B Smith 3-E -> John Smith 3
+
+    X = []
+    Y = []
+    with open(data_path) as f:
+        lines = f.read().split('ID ')
+        for line in lines:
```
nsorros, line 138: same, i think this is not lines. if not i suggest naming accordingly

wellcomeml/datasets/winer.py
```diff
@@ -0,0 +1,222 @@
...
+    random.seed(rand_seed)
+    n_train = round(prop_train*n_sample)
+
+    print("Creating entities dictionary from {}".format(NE_path))
+    entities = {}
+    with tarfile.open(NE_path, "r:bz2") as tar:
+        for member in tqdm(tar.getmembers()):
+            f = tar.extractfile(member)
+            if f:
+                content = f.read().decode('utf-8', errors='ignore')
```
nsorros, line 54: are you sure ignorning is the right behavior? don't they mention the right format somewhere?

wellcomeml/datasets/winer.py
```diff
@@ -0,0 +1,222 @@
...
+    id2word = {}
+    with open(vocab_path, "r") as vocab:
+        f = vocab.read()
+        lines = f.split('\n')
+        for i, line in enumerate(lines):
+            id2word[i] = line.split(' ')[0]
+
+    # Go into a sample of the document files and get the text for these entities
+    print("Creating token and tag from {}".format(docs_path))
+    with tarfile.open(docs_path, "r:bz2") as tar_docs, \
```
nsorros, line 86: interesting, did not know you could open three files at the same time like that

wellcomeml/datasets/winer.py
```diff
@@ -0,0 +1,222 @@
...
+            open(train_processed_path, 'w') as train_file, \
+            open(test_processed_path, 'w') as test_file:
+        for i, member in tqdm(enumerate(random.sample(tar_docs.getmembers(), n_sample))):
```
nsorros, line 89: is this to shuffle? `random.sample(tar_docs.getmembers(), n_sample)` why don't you use `random.shuffle`? also too many things in one line i suggest doing this before iterating

wellcomeml/datasets/winer.py
```diff
@@ -0,0 +1,222 @@
...
+            if i < n_train:
+                output_file = train_file
+            else:
+                output_file = test_file
+            f = tar_docs.extractfile(member)
+            if f:
```
nsorros, line 95: there are empty files?

wellcomeml/datasets/winer.py
lizgzil: oops, it happens every time actually (since the content starts with `ID`) this was just here at the beginning to test that, but I forgot to remove it

wellcomeml/datasets/winer.py
lizgzil: good point!

wellcomeml/datasets/winer.py
lizgzil: Ah yes, this should be indented, good spot! I want each sentence within a article to be separated by a newline, like
```
ID 123
The O
Liz PER
is O
. O

In O
Germany LOC
. O

ID 345
Another O
article O
```

aCampello:
Do we need the entire vocabulary in the tests? Seems like a very big file.

Screenshot 2020-06-15 at 11 22 47

lizgzil:
> Do we need the entire vocabulary in the tests? Seems like a very big file.
> 
> Screenshot 2020-06-15 at 11 22 47

@aCampello I was worried this might come up :( This is actually already the cut down vocab file for the tests, but yes if it is too big then I can work to make a more basic test dataset which needs a smaller vocab file

tests/datasets/test_winer.py
```diff
@@ -0,0 +1,92 @@
...
+    # Create the train/test data
+    n_sample = 10
+    prop_train = 0.5
+
+    create_train_test(
+        NE_path, vocab_path, docs_path,
+        train_processed_path, test_processed_path,
+        n_sample, prop_train, rand_seed=42
+        )
+    assert os.path.exists(train_processed_path) and os.path.exists(test_processed_path)
```
nsorros, line 35: possibly a better test is to test a function that creates and returns the datasets before saving and test some elements. i think it is generally suggested to avoid IO and network operations in tests unless you are explicitly testing for those i.e. a function whose purpose is solely to do IO.

nsorros:
> > Do we need the entire vocabulary in the tests? Seems like a very big file.
> > Screenshot 2020-06-15 at 11 22 47
> 
> @aCampello I was worried this might come up :( This is actually already the cut down vocab file for the tests, but yes if it is too big then I can work to make a more basic test dataset which needs a smaller vocab file

Instead of uploading files I suggest you create a tmp file with python that you can act as the file for your tests.

wellcomeml/datasets/winer.py
```diff
@@ -0,0 +1,271 @@
...
+                    entities[article_id] = sentence_entities
+
+    # Create id2word from vocab file
+    # The WiNER documentation states that the line number is the word id
+    logger.info("Creating id2word dictionary from {}".format(vocab_path))
+    id2word = {}
+    with open(vocab_path, "r") as vocab:
+        f = vocab.read()
+        lines = f.split('\n')
+        for i, line in enumerate(lines):
```
nsorros, line 104: i think you can just do `or i, line in enumerate(vocab):` and remove the above two lines

wellcomeml/datasets/winer.py
```diff
@@ -0,0 +1,271 @@
...
+    # entities = {article_id: {sentence_id:[[0, 4, 0], [18, 25, 1]]}}
+    # The sentence id is which number sentence (e.g. 1st, 2nd) the entities
+    # are from in the whole document
+
+    logger.info("Creating entities dictionary from {}".format(NE_path))
+    entities = {}
+    with tarfile.open(NE_path, "r:bz2") as tar:
+        for member in tqdm(tar.getmembers()):
+            f = tar.extractfile(member)
+            if f:
```
nsorros, line 70: to avoid nesting everything below once more you can just do
```
if f:
    continue
```
a comment as to when this happen might be useful as well

wellcomeml/datasets/winer.py
```diff
@@ -0,0 +1,271 @@
...
+                content = f.read().decode('utf-8', errors='ignore')
+                articles = content.replace('\t', ' ').split('ID ')
+                # The first element will be blank from the splitting
+                articles = articles[1:]
+                for article in articles:
```
nsorros, line 75: one yay to make this easier to follow is to create a function that goes line by line and yields article_entities when `ID` is encounter and check for empty articles. Something like
```
def yield_article_entities(f):
    article_entities = []
    for line in f:
        entities_data = line.replace('\t', '')
        if entities_data: # (worth checking if still needed)
             sentence_id, begin, end, ent_type = entities_data.split(' ')
             article_entities.append((sentence_id, begin, end, ent_type))
        if line.startswith('ID '):
            yield article_entities
     yield article_entities
```

and then you can use it here `for sentence_id, begin, end, ent_type in yield_article_entities(f):`

tests/datasets/test_winer.py
lizgzil: i will just delete this test then. the next test does as you say anyway - creates the datasets and tests some elements of them

wellcomeml/datasets/winer.py
lizgzil: oh cool, didnt think that

wellcomeml/datasets/winer.py
lizgzil: good idea, i've implemented this with a few tweaks

lizgzil:
@aCampello @nsorros latest commits - add new mock test data which is smaller (antonios comment), edit test_winer to account for new mock data, remove a os based test (nicks comment), prefilter the tar files rather than loads of 'ifs', use a yield function to make the entities creation a bit clearer (nicks comment)

wellcomeml/datasets/winer.py
```diff
@@ -0,0 +1,279 @@
...
+        for line in lines:
+            if line != '':
+                sentences = line.split('\n\n')
+                # The first and last sentence is blank due to splitting
+                sentences = sentences[1:-1]
+                for sentence in sentences:
+                    char_i = 0 # A counter to populate the start and end character indexes for each entity
+                    sentence_text = ''
+                    sentence_tags = []
+                    entities = sentence.split('\n') # ['Gibraltar 3-B', 'has O', 'a O', ...]
```
nsorros, line 199: it would possibly be easier to iterate over lines in the first place and just identify places whereas two successive newlines occured

wellcomeml/datasets/winer.py
```diff
@@ -0,0 +1,279 @@
...
+                    tokens, tags = zip(*[tuple(ee) for ee in [e.split(' ') for e in entities]])
+                    if merge_entities:
```
nsorros, line 201: it would also be more readable to go through this transformation on another function so that the logic is that one function produces tokens, tags and another takes tokens and tags and merges some together returning tokens and tags

wellcomeml/datasets/winer.py
```diff
@@ -0,0 +1,279 @@
...
+                        # if entities span over multiple words merge them into one token
+                        # e.g. John 3-B Smith 3-E -> John Smith 3
+                        prev_B = 0 # Counter for index of last entity beginning
+                        group_tags = []
+                        for i, tag in enumerate(tags):
```
nsorros, line 206: this would also be simpler if it was yielding from function grouped tags and the grouping happened in the other function

lizgzil:
@nsorros thanks nick, I've made your suggestions

PR #63, opened 2020-06-11
Add pretrained param to doc2vec

nsorros:
Adds a pretrained parameter in doc2vec that allows to load a pretrained model from a path so that the model can pretrained on a different dataset than the one that is applied which is quite common lately in NLP

### DIFF
--- a/wellcomeml/ml/doc2vec_vectorizer.py
+++ b/wellcomeml/ml/doc2vec_vectorizer.py
@@ -28,7 +28,7 @@ class Doc2VecVectorizer(BaseEstimator, TransformerMixin):
             model: underlying model architecture, one of dm or dbow. default: dm
             epochs: number of passes over training data. default: 20
             n_jobs: number of cores to use (-1 for all). default: 1
-            pretrained: pretrained doc2vec model
+            pretrained: path to directory containing saved pretrained doc2vec artifacts
         """
         self.vector_size = vector_size
         self.window_size = window_size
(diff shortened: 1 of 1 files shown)


wellcomeml/ml/doc2vec_vectorizer.py
```diff
@@ -27,6 +28,7 @@ def __init__(self, vector_size=100, window_size=5, n_jobs=1,
             model: underlying model architecture, one of dm or dbow. default: dm
             epochs: number of passes over training data. default: 20
             n_jobs: number of cores to use (-1 for all). default: 1
+            pretrained: pretrained doc2vec model
```
lizgzil, line 31: I'm not sure its obvious what is wanted here, so maybe specify this is the directory to the model, and also not the directory to the actual doc2vec file (since in _get_model_path you do `"{}/doc2vec".format(model_dir)`)

nsorros: very good point, i was expecting this comment :) i am doing this because more than one files are produced when saving. this parameter is sort of similar to `from_pretrained` that hugging face exposes from loading pretrained BERT like models. i agree though that the name is not super clear. do you have a suggestion? or you mean you happy with a better description in the param docstring?

lizgzil: Yeh I think just updating the docstring will be enough

PR #65, opened 2020-06-16
Download en_core_web_sm if not available

nsorros:
This is the same approach I have followed with sent2vec vectorizer but for a different reason. This implements the ability to download en_core_web_sm when needed. At the moment `python -c "import wellcomeml.ml"` fails upon installation because this is missing which in a way makes sense but in another is a bad user experience.


nsorros:
Actually I get a weird behaviour, when I run `python -c "import wellcomeml.ml"` it downloads and install en_core_web_sm but fails with IOError, upon rerunning it, it succeeds as if some environment variable is updated and is not picked up

Issue #45, opened 2020-05-06
Fix multi class classification in CNN and BiLSTM


nsorros:
Multi class classification loss should be categorical cross entropy and predictions should not be the neurons that are more than 50% but the argmax.

Issue #42, opened 2020-04-23
Better implementation of metrics in NN architectures


nsorros:
Instead of a callback pass a custom function that works on calculating the metrics of interest. This does not work if we naively pass f1_score or even a partial f1_score with the average instantiated for different reasons.

nsorros:
@aCampello actually there is also a ticket for the metrics, want to write down what you plan to implement if you want to work on this?

aCampello:
Yes. I was planning to:

- Define metrics as their own custom metric as in https://keras.io/metrics/
- Add a callback to add metrics to the history of model at the end of each mini-batch (instead of epoch), so we could potentially plot a history by calling `classifier.model.history`.

Something that will probably look like

`def get_keras_custom_metrics(metric_names)`

and the possible metric names are `f1_score`, `precision`, `recall`, etc, and this will return a Keras-compatible metric. So from the classifiers you could pass the parameter
`metrics=get_keras_custom_metrics(['f1_score', 'precision', 'recall'])`

You could even add built-in metrics with this to the model like:
`metrics += ['sparse_categorical_cross_entropy']`

Is that similar to what you thought?

nsorros:
I thought each metric would have its own function and get called from metrics something like `metrics=[f1_score, prediction_score, recall_score]`

nsorros:
if they are working out of the box 100%

PR #69, opened 2020-06-18
Feature/keras metrics

aCampello:
Adds Keras metrics (precision, recall, f1) and saves the metrics per-mini-batch in the model history to allow us to plot a fine evolution of training metrics.

This partially addresses https://github.com/wellcometrust/WellcomeML/issues/42. Both additions also works for the `cnn_classifier` and `bilstm_classifier` in the binary case, but I didn't want to make breaking changes so didn't turn it on by default.

Perhaps when we address https://github.com/wellcometrust/WellcomeML/issues/45, we can add the non-binary case to the metrics.

### DIFF
--- a/wellcomeml/ml/keras_utils.py
+++ b/wellcomeml/ml/keras_utils.py
@@ -1,3 +1,5 @@
+from collections import defaultdict
+
 from sklearn.metrics import f1_score, precision_score, recall_score
 import tensorflow as tf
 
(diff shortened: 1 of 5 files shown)


aCampello:
This broke the pipeline, investigating now.

wellcomeml/ml/keras_utils.py
```diff
@@ -15,3 +18,95 @@ def on_epoch_end(self, epoch, logs):
...
+
+        tp = tf.reduce_sum(
+            tf.cast(
+                tf.logical_and(greater_than_threshold, positive),
+                'float32'
+            )
+        )
+
+        if self.metric in ['precision', 'f1_score']:
+            tp_plus_fp = tf.reduce_sum(
```
nsorros, line 67: would it not be more general to calculate tp, fp, fn and calculate all metrics accordingly

aCampello: That makes sense, the idea is to add more metrics in the future, so definitely worth thinking better ways to do this. For now the different `if` statements save a bit of calculations...

Issue #67, opened 2020-06-17
Installation of en_core_web_sm and sent2vec upon usage


nsorros:
At the moment, these models/packages are downloaded first time a user is import from `wellcomeml.ml`. This is weird when a user is trying to use a package that does not need these models. It would be preferable to download them when a user imports or tries to use a class that depends on those packages/models

PR #71, opened 2020-06-18
Moving model downloads inside classes

lizgzil:
Fixes https://github.com/wellcometrust/WellcomeML/issues/67

- in `frequency_vectorizer.py` I move the `nlp` download to `spacy_lemmatizer` rather than the init since spacy_lemmatizer doesnt always need to be used and it is the only place en_core_web_sm is needed

Also
- download model when needed in spacy knowledge base too (which I think would have been breaking if i hadnt done this since en_core_web_lg was never downloaded!)

I went through the other wellcome ml files to look for any other times models were downloaded, but could spot any. But let me know if you know of any and I can make sure their download is only done when necessary.

Note for putting these checks in classes: it failed for `Sent2VecVectorizer` when the download checks were in the __init__, so I put them in the `fit` function since this is where sent2vec is needed, and then it worked

### DIFF
--- a/wellcomeml/__main__.py
+++ b/wellcomeml/__main__.py
@@ -12,7 +12,7 @@ def download(download_target):
             'python', '-m', 'spacy', 'download', 'en_trf_bertbaseuncased_lg'])
     elif download_target == "non_pypi_packages":
         subprocess.run([
-            'pip', 'install', 'git+https://github.com/epfml/sent2vec.git'])
+            'build/virtualenv/bin/pip', 'install', 'git+https://github.com/epfml/sent2vec.git'])
     else:
         print(f"{download_target} is not one of models,deeplearning-models")
 
--- a/wellcomeml/__main__.py
+++ b/wellcomeml/__main__.py
@@ -12,7 +12,7 @@ def download(download_target):
             'python', '-m', 'spacy', 'download', 'en_trf_bertbaseuncased_lg'])
     elif download_target == "non_pypi_packages":
         subprocess.run([
-            'build/virtualenv/bin/pip', 'install', 'git+https://github.com/epfml/sent2vec.git'])
+            'pip', 'install', 'git+https://github.com/epfml/sent2vec.git'])
     else:
         print(f"{download_target} is not one of models,deeplearning-models")
 
--- a/wellcomeml/ml/sent2vec_vectorizer.py
+++ b/wellcomeml/ml/sent2vec_vectorizer.py
@@ -8,7 +8,11 @@ from wellcomeml.utils import check_cache_and_download
 
 
 class Sent2VecVectorizer(BaseEstimator, TransformerMixin):
+
     def __init__(self, pretrained=None):
+        self.pretrained=pretrained
+
+    def fit(self, *_):
 
         try:
             import sent2vec
@@ -16,10 +20,7 @@ class Sent2VecVectorizer(BaseEstimator, TransformerMixin):
             from wellcomeml.__main__ import download
             download("non_pypi_packages")
             import sent2vec
-    
-        self.pretrained=pretrained
 
-    def fit(self, *_):
         if self.pretrained:
             model_path = check_cache_and_download(self.pretrained)
             self.model = sent2vec.Sent2vecModel()
--- a/wellcomeml/ml/sent2vec_vectorizer.py
+++ b/wellcomeml/ml/sent2vec_vectorizer.py
@@ -3,18 +3,20 @@ Vectorizer that exposes sklearn interface to sent2vec
 paper and codebase. https://github.com/epfml/sent2vec
 """
 from sklearn.base import TransformerMixin, BaseEstimator
-try:
-    import sent2vec
-except ImportError:
-    from wellcomeml.__main__ import download
-    download("non_pypi_packages")
-    import sent2vec
 
 from wellcomeml.utils import check_cache_and_download
 
 
 class Sent2VecVectorizer(BaseEstimator, TransformerMixin):
     def __init__(self, pretrained=None):
+
+        try:
+            import sent2vec
+        except ImportError:
+            from wellcomeml.__main__ import download
+            download("non_pypi_packages")
+            import sent2vec
+    
         self.pretrained=pretrained
 
     def fit(self, *_):
--- a/wellcomeml/ml/frequency_vectorizer.py
+++ b/wellcomeml/ml/frequency_vectorizer.py
@@ -11,17 +11,6 @@ import spacy
 from sklearn.feature_extraction.text import TfidfVectorizer
 
 from wellcomeml.logger import logger
-try:
-    nlp = spacy.load('en_core_web_sm', disable=['ner', 'tagger', 'parser',
-                                                'textcat'])
-except IOError:
-    from wellcomeml.__main__ import download
-    download("models")
-    # pkg_resources need to be reloaded to pick up the newly installed models
-    import pkg_resources, imp
-    imp.reload(pkg_resources)
-    nlp = spacy.load('en_core_web_sm', disable=['ner', 'tagger', 'parser',
-                                            'textcat'])
 
 
 class WellcomeTfidf(TfidfVectorizer):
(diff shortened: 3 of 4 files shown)


lizgzil:
it also feels like we have a duplication of effort with 
```
$(VIRTUALENV)/.models:
	$(VIRTUALENV)/bin/python -m spacy download en_core_web_sm
	touch $@

$(VIRTUALENV)/.deep_learning_models:
	$(VIRTUALENV)/bin/python -m spacy download en_trf_bertbaseuncased_lg
	touch $@

$(VIRTUALENV)/.non_pypi_packages:
	$(VIRTUALENV)/bin/pip install git+https://github.com/epfml/sent2vec.git
	touch $@
```
in the `Makefile`, and
```
if download_target == "models":
        subprocess.run([
            'python', '-m', 'spacy', 'download', 'en_core_web_sm'])
    elif download_target == "deeplearning-models":
        subprocess.run([
            'python', '-m', 'spacy', 'download', 'en_trf_bertbaseuncased_lg'])
    elif download_target == "non_pypi_packages":
        subprocess.run([
            'pip', 'install', 'git+https://github.com/epfml/sent2vec.git'])
    else:
        print(f"{download_target} is not one of models,deeplearning-models")
```
in the `__main__.py`.

And at the moment there is only one thing to download for each `download_target` option, but in the future we might want to just be specific when we download? e.g. what if `download("non_pypi_packages")` downloaded loads of things but we only wanted sent2vec?

Any ideas to make this neater/am I missing something and we should have duplication in both the `Makefile` and `__main__.py`?

nsorros:
> it also feels like we have a duplication of effort with
> 
> ```
> $(VIRTUALENV)/.models:
> 	$(VIRTUALENV)/bin/python -m spacy download en_core_web_sm
> 	touch $@
> 
> $(VIRTUALENV)/.deep_learning_models:
> 	$(VIRTUALENV)/bin/python -m spacy download en_trf_bertbaseuncased_lg
> 	touch $@
> 
> $(VIRTUALENV)/.non_pypi_packages:
> 	$(VIRTUALENV)/bin/pip install git+https://github.com/epfml/sent2vec.git
> 	touch $@
> ```
> 
> in the `Makefile`, and
> 
> ```
> if download_target == "models":
>         subprocess.run([
>             'python', '-m', 'spacy', 'download', 'en_core_web_sm'])
>     elif download_target == "deeplearning-models":
>         subprocess.run([
>             'python', '-m', 'spacy', 'download', 'en_trf_bertbaseuncased_lg'])
>     elif download_target == "non_pypi_packages":
>         subprocess.run([
>             'pip', 'install', 'git+https://github.com/epfml/sent2vec.git'])
>     else:
>         print(f"{download_target} is not one of models,deeplearning-models")
> ```
> 
> in the `__main__.py`.
> 
> And at the moment there is only one thing to download for each `download_target` option, but in the future we might want to just be specific when we download? e.g. what if `download("non_pypi_packages")` downloaded loads of things but we only wanted sent2vec?
> 
> Any ideas to make this neater/am I missing something and we should have duplication in both the `Makefile` and `__main__.py`?

The idea here, and you can definitely and should open an issue about it, is that a user that installs the basic library would have one command to download all models needed and similarly for someone that downloads the deep learning version there is another command that downloads the deep learning dependencies so we maintain these two commands and add models that are needed for each installation. The other command is not about models, that is why i created a separate command, it is about packages that are not on pypi so they can only be installed manually. An altertnative is definitely to be explicity which we can implement, if so we need to throw explicit errors to the user to tell them what to download as they should not be obliged to know all the models and how to call download

nsorros:
> Screenshot 2020-06-18 at 17 41 27
> 😢

could it be because you instantiate inside the init instead of somewhere in the class randomly as you do for spacy?
```
class Sent2Vec:
    try:
        import sent2vec
    ...
    def __init__():
```

nsorros:
That is a great start @lizgzil, thanks. If you resolve the test, this is ready to merge, very excited, super helpful.

wellcomeml/__main__.py
```diff
@@ -12,7 +12,7 @@ def download(download_target):
             'python', '-m', 'spacy', 'download', 'en_trf_bertbaseuncased_lg'])
     elif download_target == "non_pypi_packages":
         subprocess.run([
-            'pip', 'install', 'git+https://github.com/epfml/sent2vec.git'])
+            'build/virtualenv/bin/pip', 'install', 'git+https://github.com/epfml/sent2vec.git'])
```
nsorros, line 15: that assumes that there is a `build/virtualenv` in the user, a user might call their virtualenv differently, for example as i did in the demo `python3 -m venv venv`. i did the same mistake when first implemented it but eventually we decided with @aCampello that is better to delegate in which virtualenv to install to the user

aCampello: Oh yeah, that's true. `pip` would be preferable. I missed this from the review!

wellcomeml/ml/sent2vec_vectorizer.py
```diff
@@ -3,21 +3,24 @@
...
 
 
 class Sent2VecVectorizer(BaseEstimator, TransformerMixin):
+
     def __init__(self, pretrained=None):
         self.pretrained=pretrained
 
     def fit(self, *_):
+
+        try:
```
nsorros, line 17: why inside fit and not outside? what happens if a user pickles the model and then loads and runs transform? i suspect that sent2vec will not be imported and it will fail, if I am right it is worth adding a test

lizgzil: it failed the tests when it was in `__init__` (sent2vec not found). At the moment it looks like you can't load a model in this class anyway, so they'd have to fit before transforming anyway. 

I think that if a `load_model` function was added to this class, this is where the import sent2vec lines would live.

wellcomeml/__main__.py
lizgzil: hm good point, but when it was just pip install the tests failed (sent2vec not found). So it does seem to be neccessary to pip install it to a particular location

lizgzil: (and these are the tests as run on Travis not my machine)

lizgzil: is there a way we can do this systematically? i.e. know which `build/virtualenv/bin/pip` or equivalent location to install to?

wellcomeml/ml/sent2vec_vectorizer.py
nsorros: i am worried that when the class is used after pickling it will fail. i was suggesting that you try to add it before init as you have done everywhere else. let me give it a try to see if fails

wellcomeml/__main__.py
nsorros: no, this is the responsibility of the user, this should not fail at the tests as the virtualenv should be enabled

wellcomeml/ml/sent2vec_vectorizer.py
nsorros: actually, pickling fails atm anyway so no need to worry about that

lizgzil:
@nsorros in response to your comments:

My process of debugging for these 2 things (where to download sent2vec within `sent2vec_vectorizer.py` and how to install it in `__main__.py`) was:

commit 1. download in `init` + `pip` install -> failed in Travis
commit 2. download in `fit` + `pip` install -> failed in Travis
commit 3. download in `fit` + `build/virtualenv/bin/pip` install -> succeeded in Travis
commit 4. download in `init` + `build/virtualenv/bin/pip` install -> failed in Travis
step 5. (removed commit 4)

[note: I cant test locally since I can't install sent2vec anyway, so I had to debug by step by step commits and checking the Travis checks.]

but what you are saying that the next experiment is:
download at the beginning of the class not in a function + `pip` install

nsorros:
> [note: I cant test locally since I can't install sent2vec anyway, so I had to debug by step by step commits and checking the Travis checks.]
Note that you can run tests on your AWS instance

> but what you are saying that the next experiment is:
> download at the beginning of the class not in a function + `pip` install
yes, that might be it but if not you need to better understand why it was succeeding before and failing now

Makefile
```diff
@@ -10,7 +10,6 @@ $(VIRTUALENV)/.installed:
 	$(VIRTUALENV)/bin/pip3 install -r requirements.txt
 	$(VIRTUALENV)/bin/pip3 install -r requirements_test.txt
 	$(VIRTUALENV)/bin/pip3 install -e .
-	$(VIRTUALENV)/bin/pip3 install git+https://github.com/epfml/sent2vec.git # otherwise it fails cause setup.py relies on cython and numpy being installed
```
nsorros, line 13: the problem is caused from removing this line. adding it should allow you remove the need for `build/virtuelanv/bin/pip`. this is not a good permanent fix but i am working on a better solution

lizgzil: cool. Maybe it makes sense to leave this PR until you've merged your PR and then I'll rebase on it?

nsorros:
> Fixing #67

To automatically close / link an issue you need to say either fixed, or fixes or fix https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

nsorros:
Did the change and rebased to make sure the change is not overwritten. If tests pass we can merge.

lizgzil:
@nsorros woop! thanks for sorting

Issue #33, opened 2020-04-16
Add native HuggingFace Scibert support


aCampello:
In light of https://github.com/allenai/scibert/commit/7598219a8d80b9c2fe1323a141e4a9e40ec044cb, and the closed issue https://github.com/allenai/scibert/issues/71, we don't need cache for Scibert anymore.

It could still be useful to have the "download_and_cache" function if we need another model, but for the time being we can turn it off for Scibert.

nsorros:
I think this is slightly more complicated as we do not use hugging face directly but spacy transformers. We should of course move towards hugging face slowly as this will give us more control but for the time being I am not sure if we can leverage the integration. Need to investigate.

aCampello:
I think some parts of the code, including the `BertVectorizer` and the `SemanticSimilarityClassifier` can already use the native hugging face. The only blocker is space transformers. In their next release, they are likely to support new versions of transformers (https://github.com/explosion/spacy-transformers/blob/bf729718502053ac9bd8770fc67aaeb549593510/requirements.txt#L2), so after that we can probably try to tackle this issue again.

aCampello:
I am hoping this will be doable with the new spacy-transformers.

PR #76, opened 2020-06-22
Pin spacy transformers version

### LINKED ISSUE #75: Spacy-transformers incompatible with transformers
# Problem
At the moment when installing WellcomeML, the latest version of spacy transformers is installed which creates a conflict with the transformers requirement of below 2.
```
ERROR: spacy-transformers 0.6.1 has requirement spacy<2.4.0,>=2.3.0, but you'll have spacy 2.2.1 which is incompatible.
ERROR: spacy-transformers 0.6.1 has requirement transformers<2.9.0,>=2.4.0, but you'll have transformers 2.0.0 which is incompatible.
ERROR: tensorflow 2.2.0 has requirement scipy==1.4.1; python_version >= "3", but you'll have scipy 1.5.0 which is incompatible.
```

# Replicating
You can replicate by installing the latest version of WellcomeML with deep learning. Downloading the models (which for some reason it does not work automatically, possibly another issue) and try to import wellcomeml.ml
```
python3 -m venv venv
source venv/bin/activate
pip install wellcomeml[deep-learning]
python -m wellcomeml download models
python -c "import wellcomeml.ml"
```

this throws
```
pkg_resources.VersionConflict: (transformers 2.0.0 (/Users/nicksorros/code/test/venv/lib/python3.7/site-packages), Requirement.parse('transformers<2.9.0,>=2.4.0'))
```

Interestingly the tests do not fail since when creating the virtualenv we do not install the deep learning requirements from the setup.py but the requirements.txt.

# Solution
A temporary fix is to pin the spacy transformers library to 0.5.1. An alternative is to ensure the library works with the latest transformers release.

Additionally when creating the virtualenv we should install the deep learning requirements from the setup.py `pip install -e .[deep-learning]`. 

Lastly, maintaining a requirements.txt and a setup.py creates an inconsistency since when a user installs the dependencies are defined by setup.py but when we develop or run tests we use mostly the requirements (as this runs first anyway). So removing requirements.txt all together will resolve future bugs similar to this. An alternative would be to read the requirements from setup.py but separating them to extras vs no extras could be a bit painful

nsorros:
Fixes #75 where the problem is described in more detail.

TL;DR. This PR removes requirements.txt and pins the version of spacy transformers. The first ensures that similar failures can be cached from tests and the second resolves a version conflict that the latest release of spacy-transformers introduced


aCampello:
I am happy with only having setup.py, but I we might need to include other packages explicitly to make it work, for example `Cython` (see pipeline logs)

Makefile
```diff
@@ -7,30 +7,13 @@ $(VIRTUALENV)/.installed:
 	@if [ -d $(VIRTUALENV) ]; then rm -rf $(VIRTUALENV); fi
 	@mkdir -p $(VIRTUALENV)
 	virtualenv --python $(PYTHON_VERSION) $(VIRTUALENV)
-	$(VIRTUALENV)/bin/pip3 install -r requirements.txt
 	$(VIRTUALENV)/bin/pip3 install -r requirements_test.txt
-	$(VIRTUALENV)/bin/pip3 install -e .
-	$(VIRTUALENV)/bin/pip3 install git+https://github.com/epfml/sent2vec.git # otherwise it fails cause setup.py relies on cython and numpy being installed
```
nsorros, line 13: I think the problem @aCampello was caused by removing this line not requirements. Testing that theory now.

aCampello: I think the problem is that when we download sent2vec it doesn't install the dependencies of that project.

nsorros: Issue #83 and resolved by #84

aCampello: 👍

aCampello:
Does this also solve  https://github.com/wellcometrust/WellcomeML/issues/33?

nsorros:
> Does this also solve #33?

How so?

aCampello:
Nevermind, I thought the new version of transformers somehow installed things on the fly as well. But I think we can resolve that issue quite soon anyway.

Issue #62, opened 2020-06-10
Improve docs


aCampello:
Would be good to improve docs in order to have a list of:

- [ ] All vetorisers/classifiers 
- [ ] The download functionality

PR #78, opened 2020-06-22
Add documentation

aCampello:
Adds documentation webpage to the project.

At the moment this is being served from this branch (`gh-pages`) at http://ghp.wellcomecollection.org/WellcomeML/docs/_build/html/index.html. We will have to change the domain name to `wellcomedatalabs`. 

After this branch is merged, I will switch the docs to generate from `master/docs` instead.

Subsequently, docs will be generated automatically with new releases, and/or can be changed at any time by following the new procedure in the README.md.

### 2.2 Contributing to the docs

Make changes to the .rst files in `/docs` (please **do not change the ones starting by wellcomeml** as those are generated automatically)

Navigate to the root repository and run

```bash
make update-docs
```

Closes https://github.com/wellcometrust/WellcomeML/issues/62


Makefile
```diff
@@ -9,9 +9,15 @@ $(VIRTUALENV)/.installed:
 	virtualenv --python $(PYTHON_VERSION) $(VIRTUALENV)
 	$(VIRTUALENV)/bin/pip3 install -r requirements.txt
 	$(VIRTUALENV)/bin/pip3 install -r requirements_test.txt
+	$(VIRTUALENV)/bin/pip3 install -r docs/requirements.txt # Installs requirements to docs
 	$(VIRTUALENV)/bin/pip3 install -e .
 	touch $@
 
+.PHONY: update-docs
+update-docs:
+	cd docs && ../$(VIRTUALENV)/bin/sphinx-apidoc --no-toc -d 5 -H WellcomeML -o . -f ../wellcomeml
```
nsorros, line 18: can this not run without cd ing?

aCampello: You are right, we can. I changed that in the new commit. The next command though depends on a makefile in the sub-directory so to run without cd it would require bigger changes, so I left it like that.

PR #87, opened 2020-06-24
Add SemanticMetaBert

aCampello:
Adds a class `SemanticMetaBert` that extends `SemanticSimilarityClassifier` to enable numerical metadata to be added to the model. The new model follows the following diagram:

Screenshot 2020-06-24 at 12 34 08

I will look into adding `BatchNorm` (or `LayerNorm`) in the future, but for the time being will assume that the user will scale the numerical features correctly before feeding the model.


aCampello:
The travis-ci seem to run out of memory to run the test, will investigate

```bash
<string>:3: tensorflow.python.framework.errors_impl.ResourceExhaustedError:  OOM when allocating tensor with shape[31116,768] and type float on /job:localhost/replica:0/task:0/device:CPU:0 by allocator cpu
808
809
```
Tests pass locally

wellcomeml/ml/bert_semantic_equivalence.py
```diff
@@ -198,15 +221,191 @@ def predict(self, X):
...
         self.model.save_pretrained(path)
 
     def load(self, path):
+        """Loads model from path"""
         self._initialise_models()
         self.model = TFBertForSequenceClassification.from_pretrained(path)
         self.trained_ = True
 
 
+class SemanticMetaBert(SemanticEquivalenceClassifier):
```
nsorros, line 235: To keep up with the convention of sklearn and for consistency maybe the class should be called SemanticEquivalenceMetaBertClassifier. I also think Bert is redundant in the name

wellcomeml/ml/bert_semantic_equivalence.py
```diff
@@ -198,15 +221,191 @@ def predict(self, X):
...
+
+        # Drop out layer to the concatenated features
+        x = tf.keras.layers.Dropout(rate=self.dropout_rate,
+                                    name='concatenated_dropout')(x)
+
+        # Dense layer that will be used for softmax prediction later
+        x = tf.keras.layers.Dense(2, name='dense_layer')(x)
+
+        self.meta_model = tf.keras.Model(
+            input_text_tensors + [input_numerical_data_tensor],
```
nsorros, line 298: why the need for the parenthesis in the second? maybe it would be better for those variables to have the same format from the beggining

wellcomeml/ml/bert_semantic_equivalence.py
```diff
@@ -198,15 +221,191 @@ def predict(self, X):
...
+            remaining components are numerical features (e.g.)
+            y: list of 0 (not similar) and 1s (similar)
+            random_state: a random state for the train_test_split
+            epochs: number_of epochs
+
+        Returns:
+             Array of probabilities, of shape len(X) x 2
+
+        """
+        return super().fit(X, y, **kwargs)
```
nsorros, line 355: nice abstraction

wellcomeml/ml/bert_semantic_equivalence.py
```diff
@@ -198,15 +221,191 @@ def predict(self, X):
...
+            2+n_numerical_features, the first two components are texts and
+            remaining components are numerical features (e.g.)
+
+        Returns:
+            Array of predictions of length len(X)
+
+        """
+        return super().predict(X)
+
+    def save(self, path):
```
nsorros, line 397: isn't the below functions the same for the two models? if all the changes is the name of the variable then i suggest you investigate whether the variable holding the model does not change, same as in sklearn it almost always clf, here it can be model

wellcomeml/ml/bert_semantic_equivalence.py
```diff
@@ -198,15 +221,191 @@ def predict(self, X):
...
+        """Saves meta model to path"""
+        os.makedirs(path, exist_ok=True)
+        self.meta_model.save(path)
+
+    def load(self, path):
+        """Loads metamodel from path"""
+        self._initialise_models()
+        self.meta_model = tf.keras.models.load_model(path)
+        self.trained_ = True
```
nsorros, line 406: nice ;)

wellcomeml/ml/bert_semantic_equivalence.py
```diff
@@ -198,15 +221,191 @@ def predict(self, X):
...
+    def _prep_dataset_generator(self, X, y):
+        """Overrides/extends the super class data preparation"""
+        text_features = {"input_ids", "attention_mask", "token_type_ids"}
+        X_text, X_numerical = self._separate_features(X)
+
+        batch_encoding_text = self.tokenizer.batch_encode_plus(
+            X_text, max_length=self.max_length, add_special_tokens=True,
+        )
+
+        def gen_train():
```
nsorros, line 311: it is better to pass X, y and batch_encoding_text rather than rely python finding those variables one scope up from the function

nsorros:
> The travis-ci seem to run out of memory to run the test, will investigate
> 
> ```shell
> <string>:3: tensorflow.python.framework.errors_impl.ResourceExhaustedError:  OOM when allocating tensor with shape[31116,768] and type float on /job:localhost/replica:0/task:0/device:CPU:0 by allocator cpu
> 808
> 809
> ```
> 
> Tests pass locally

Generally tests are quite heavy at the moment, it takes almost 1 hour to run them. Should we create an issue to make them more light?

aCampello:
> > The travis-ci seem to run out of memory to run the test, will investigate
> > ```shell
> > <string>:3: tensorflow.python.framework.errors_impl.ResourceExhaustedError:  OOM when allocating tensor with shape[31116,768] and type float on /job:localhost/replica:0/task:0/device:CPU:0 by allocator cpu
> > 808
> > 809
> > ```
> > 
> > 
> > Tests pass locally
> 
> Generally tests are quite heavy at the moment, it takes almost 1 hour to run them. Should we create an issue to make them more light?

Yes, I think this is a good idea. I made the first test run, but no matter what I did yesterday I couldn't get the save/load to run on Travis.

wellcomeml/ml/bert_semantic_equivalence.py
aCampello: I agree, but this seems to be a limitation of the tensor flow `from_generator` method (https://www.tensorflow.org/api_docs/python/tf/data/Dataset#from_generator)

wellcomeml/ml/bert_semantic_equivalence.py
aCampello: +1: this was now changed

wellcomeml/ml/bert_semantic_equivalence.py
aCampello: Agree - I changed this in https://github.com/wellcometrust/WellcomeML/pull/87/commits/36f5fdfc4f37ae3c79a415b96b903f964daf586f

wellcomeml/ml/bert_semantic_equivalence.py
aCampello: Unfortunately hugging face uses `save_pretrained` while keras use `save` and `load_model` which breaks the abstraction.

I do agree though that calling the main model `self.model` is better consistently and made this change in https://github.com/wellcometrust/WellcomeML/pull/87/commits/e21169377c2d22b766e5da9abe1d91a3641d9f44

wellcomeml/ml/bert_semantic_equivalence.py
```diff
@@ -198,15 +221,191 @@ def predict(self, X):
...
+
+        Returns:
+            Numpy array of probabilities, of shape len(X) x 2
+
+        """
+
+        X_text, X_numerical = self._separate_features(X)
+
+        X_processed = self._tokenize(X_text)
+        X_processed['numerical_metadata'] = tf.convert_to_tensor(X_numerical)
```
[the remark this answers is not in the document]
aCampello: Good idea. Created in https://github.com/wellcometrust/WellcomeML/pull/87/commits/5815135ce4fddec01941650c24b3eace29e92e0f

PR #90, opened 2020-07-01
Fix to load winer NER data

lizgzil:
Fixes https://github.com/wellcometrust/WellcomeML/issues/91

Small fix that I noticed when I tried to use the data.

It didn't come up in the test since I the test was for the first data point (which this bug didn't effect), so adding a test for a later data point instead.

### DIFF
--- a/wellcomeml/datasets/winer.py
+++ b/wellcomeml/datasets/winer.py
@@ -216,9 +216,9 @@ def _load_data_spacy(data_path, inc_outside=True, merge_entities=True):
     X = []
     Y = []
     sentence_text = None
+    char_i = 0 # A counter to populate the start and end character indexes for each entity
     with open(data_path) as f:
         for line in f:
-            char_i = 0 # A counter to populate the start and end character indexes for each entity
             line = line.replace('\n', '')
             if line == '' or line[0:2] == 'ID':
                 # You are at the start of a new sentence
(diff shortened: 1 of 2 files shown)


lizgzil:
🤦 test_sent2vec.py failed

aCampello:
> 🤦 test_sent2vec.py failed

Should we organise a war cabinet to fix that - it's blocking my other PR too :(

nsorros:
> > 🤦 test_sent2vec.py failed
> 
> Should we organise a war cabinet to fix that - it's blocking my other PR too :(



> > 🤦 test_sent2vec.py failed
> 
> Should we organise a war cabinet to fix that - it's blocking my other PR too :(

happy to do that, bad sent2vec

wellcomeml/datasets/winer.py
```diff
@@ -216,8 +216,8 @@ def _load_data_spacy(data_path, inc_outside=True, merge_entities=True):
     X = []
     Y = []
     sentence_text = None
+    char_i = 0 # A counter to populate the start and end character indexes for each entity
```
nsorros, line 219: char_i is not a good name, hence the comment. ent_start_index could be a better one

aCampello:
If you rebase to master the tests should pass (just merged https://github.com/wellcometrust/WellcomeML/pull/92 that skips the problematic test)

Issue #89, opened 2020-06-30
Tests take too long


aCampello:
The scope of this issue includes:

- [x] Profiling the tests to check which ones take long
- [ ] Reducing particularly long tests

aCampello:
@nsorros

aCampello:
Here is a local profile of the tests. It takes roughly 4 minutes in my machine, but much longer in travis presumably because it needs to build the virtualenv. I am adding profiling code to the travis-ci command to see how it behaves there.

```
36.81s call     tests/test_bert_classifier.py::test_partial_fit
31.73s call     tests/test_bert_vectorizer.py::test_save_and_load
18.91s call     tests/test_bert_vectorizer.py::test_embed_long_sentence
15.72s call     tests/test_sent2vec.py::test_fit_transform
15.34s call     tests/test_bert_vectorizer.py::test_embed_one_sentence
14.81s call     tests/test_bert_vectorizer.py::test_embed_scibert
14.59s call     tests/test_bert_vectorizer.py::test_embed_two_sentences
12.78s call     tests/test_bilstm.py::test_vanilla
10.84s call     tests/test_bilstm.py::test_multilabel
10.82s call     tests/test_bert_classifier.py::test_multilabel
4.43s call     tests/test_cnn.py::test_multilabel
3.99s call     tests/test_vectorizer.py::test_bert_dispatch
3.64s call     tests/test_cnn.py::test_vanilla
1.91s call     tests/test_ner_spacy.py::test_fit
1.42s call     tests/test_vectorizer.py::test_tf_idf_dispatch
1.11s call     tests/test_vectorizer.py::test_vectorizer_that_does_not_have_save
0.81s setup    tests/test_spacy.py::test_SpacyDocToProdigy
0.76s call     tests/test_ner_spacy.py::test_score
0.65s call     tests/test_ner_spacy.py::test_ner_classification_report
0.63s call     tests/test_ner_spacy.py::test_predict
0.36s call     tests/test_spacy_classifier.py::test_multilabel
0.35s call     tests/test_spacy_classifier.py::test_multilabel_Y_list
0.28s call     tests/test_spacy_classifier.py::test_partial_fit
0.11s call     tests/test_s3_policy_data.py::test_get_keys
0.09s call     tests/test_doc2vec.py::test_score
0.04s setup    tests/test_bert_vectorizer.py::test_save_and_load
0.04s call     tests/test_entity_linking.py::test_optimise_threshold
[...]

PR #92, opened 2020-07-01
Fix/skip sent 2 vec test

aCampello:
This is a workaround https://github.com/wellcometrust/WellcomeML/issues/89 to enable our pipelines to pass until we fix the speed of the tests.

It also adds profiles to test so we can partially address https://github.com/wellcometrust/WellcomeML/issues/89.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 .DEFAULT_GOAL := all
 
 VIRTUALENV := build/virtualenv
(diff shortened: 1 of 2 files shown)


aCampello:
@nsorros are you ok with this temporary workaround? It will decrease overall coverage by just 0.32%. We can still run the skipped test locally.

PR #93, opened 2020-07-01
Feature/travis ci improvements

aCampello:


aCampello:
I don't think this will be concluded, so closing this PR for now.

PR #95, opened 2020-07-02
Run codecov from make venv than travis

### LINKED ISSUE #94: Codecov is installed in Travis venv instead of build/virtualenv make creates
When running codecov through
```
after_success:
  - codecov
```
travis looks into its own venv to find the tool.

In order to ensure the tool is available there we reinstall the rest requirements in the travis venv
```
install:
  - make virtualenv
  - pip install -r requirements_test.txt
```

This behaviour can lead to bugs that will be difficult to resolve. Ideally we want the tests and commands to run through the virtualenv that make creates so that we can debug and reason about it more easily. One possible fix is to create a command or running codecov from the our venv installation instead of running codecov which will look firstly look into the travis venv

nsorros:
Fixes #94

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,6 @@ python:
 
 install:
   - make virtualenv
-  - pip install -r requirements_test.txt
 
 script:
   - make test
@@ -24,4 +23,4 @@ branches:
   - master
 
 after_success:
-  - codecov
+  - make run_codecov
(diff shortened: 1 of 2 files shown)


aCampello:
The point where it fails has been fixed in a recent PR. If you rebase to master it should pass.

nsorros:
> The point where it fails has been fixed in a recent PR. If you rebase to master it should pass.

👍 Just did, thanks for pointing this out. Fingers crossed

nsorros:
Tests passed @aCampello, maybe you could have a look now

PR #98, opened 2020-07-08
Add workflow to run flake8

### LINKED ISSUE #97: Code style
At the moment we do not enforce any code style even though we loosely follow the [Google python style code ](https://google.github.io/styleguide/pyguide.html). 

The most commonly used code style revolves around pep8/flake8. There are two different workflows we can follow

- install some way of nagging us towards conforming to pep8/flake8
- install an automatic code formatter like [black](https://github.com/psf/black) that follows the above

The benefit of the first is that over time you learn to write better code and not produce corrections according to pep8 / flake8. The benefit of the second is that formatting is automated and taken care of.

If we choose the first option we have two different means of enforcing it.

1. Adding a check in Github similar to build and codecov for example [this](https://github.com/marketplace/actions/run-flake8-on-your-pr-with-annotations).
2. Adding a [pre commit hook](https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks) that checks and does not allow you to commit code if it does not conform.

The second is quite annoying so I would be in favour of the first if possible. I am indifferent on whether we use black or pep8 /flake8 explicitly. Let me know what you think.

nsorros:
Description
---
Fixes #97 by adding a workflow that runs flake8 and adds annotations https://github.com/marketplace/actions/flake8-action

Checklist
---

- [ ] Added tests

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,2 @@
+[flake8]
+max-line-length = 99
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,33 @@
+# This is a basic workflow to help you get started with Actions
+
+name: CI
+
+# Controls when the action will run. Triggers the workflow on push or pull request
+# events but only for the master branch
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  flake8_check:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Setup Python
+        uses: actions/setup-python@v1
+        with:
+          python-version: 3.7.8
+          architecture: x64
+      - name: Checkout Master
+        uses: actions/checkout@master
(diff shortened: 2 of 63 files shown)


nsorros:
Started fixing flake8 violations from fixing the ml folder, first by running black and then by manually correcting the rest. If you want to review only flake8 addition you can see the first commit 504c1a6f660e8e25322fcc22fedec04b13b91f55

nsorros:
> Looks good so far.
> Some notes on what seem to be common things we forget to do (might be useful):
> 
> * `"` instead of `'`
> * commas at the end of argument lists
> * brackets instead of `\` if you want to go on a new line
> * more spaces between things, e.g. `x[1 : 2 + 5]` and `[0] * (pad_len - len(x))`
> * more separating out the contents on brackets on separate lines if long
> * 2 blank lines after imports
> * double space between code and comment if on same line `x = 5  # a comment`
> * one space between hash and comment
> * classes don't need empty brackets, e.g. `class ClassName:` not `class ClassName():`

Quite useful, thanks for sharing.

PR #109, opened 2020-08-19
Add proposed clustering class

aCampello:
Description
---
Moving some functionality currently in `datalabs` to a class in Wellcome ml. Includes text clustering and some custom hyperparameter search with constraints 

Trello: <blockquote class="trello-card"><a href="https://trello.com/c/qnoMaee9/1061-add-a-clustering-class-to-wellcomeml">Add a clustering class to WellcomeML</a></blockquote>

Checklist
---

- [x] Added tests

### DIFF
--- a/wellcomeml/ml/clustering.py
+++ b/wellcomeml/ml/clustering.py
@@ -6,6 +6,8 @@ from sklearn.manifold import TSNE
 from sklearn.model_selection import ParameterGrid
 from sklearn.metrics import silhouette_score
 from sklearn.cluster import DBSCAN, OPTICS, KMeans
+from sklearn.base import ClusterMixin
+from hdbscan import HDBSCAN
 import umap
 
 from wellcomeml.ml import Vectorizer
@@ -32,7 +34,8 @@ class TextClustering(object):
         Args:
             embedding(str): 'tf-idf', 'bert', 'keras' or 'doc2vec'.
             reducer(str): 'umap' or 'tsne'
-            clustering(str): Only 'dbscan' allowed for now
+            clustering(str): 'dbscan', 'kmeans', 'optics', 'hdbscan' or any
+            class that is a sklearn.base.ClusterMixin
             cluster_reduced(bool): Whether clustering will be applied after
             reducing the datapoints with reducer, or straight after embedding
             n_kw(int): If n_kw > 0 and embedding='tf-idf', the model
@@ -58,12 +61,18 @@ class TextClustering(object):
             'dbscan': DBSCAN,
             'kmeans': KMeans,
             'optics': OPTICS,
+            'hdbscan': HDBSCAN
         }
 
-        self.clustering = clustering_dispatcher[clustering](
-            **params.get('clustering', {})
-        )
-
+        if clustering in clustering_dispatcher:
+            self.clustering = clustering_dispatcher[clustering](
+                **params.get('clustering', {})
+            )
+        elif isinstance(clustering, ClusterMixin):
+            self.clustering = clustering(**params.get('clustering', {}))
+        else:
+            raise ValueError('clustering has to be one of the available '
+                             'clusters or a sklearn ClusterMixin.')
         self.cluster_ids = None
         self.cluster_names = None
         self.cluster_kws = None
--- a/wellcomeml/ml/clustering.py
+++ b/wellcomeml/ml/clustering.py
@@ -1,14 +1,29 @@
-import umap
+from copy import deepcopy
 
+import numpy as np
 from sklearn.manifold import TSNE
+from sklearn.model_selection import ParameterGrid
+from sklearn.metrics import silhouette_score
 from sklearn.cluster import DBSCAN
+import umap
 
 from wellcomeml.ml import Vectorizer
 
 
-class ClusterPipeline(object):
+class TextClustering(object):
+    """
+    Class
+
+    Attributes:
+        vectorizer: The embedding Vectorizer object
+        reducer: A dimensionality reduction object
+        clustering: A clustering model object
+        cluster_ids: Ids of the cluster
+        cluster_names: Names of the clusters
+        cluster_kws: Keywords for the clusters (only if embedding=tf-idf)
+    """
     def __init__(self, embedding='tf-idf', reducer='umap', clustering='dbscan',
-                 params={}):
+                 n_kw=10, params={}):
         """
         Initialises a cluster pipeline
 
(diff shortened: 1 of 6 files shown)


wellcomeml/ml/clustering.py
```diff
@@ -0,0 +1,321 @@
...
+        logger.setLevel(logging_level)
+
+        return {
+            "params_list": params_list,
+            "silhouette": best_silhouette,
+            "best_params": best_params
+        }
+
+    def stability(self):
+        """Function to calculate how stable the clusters are"""
```
aCampello, line 205: FYI @nonieA , will look into implementing stability in the future. Happy to have your contribution here if you want.

wellcomeml/ml/clustering.py
```diff
@@ -0,0 +1,321 @@
...
+        Args:
+            X: A list of texts
+
+        Returns:
+            A TextClustering object
+
+        """
+        self._fit_step(X, step='vectorizer')
+        self._fit_step(step='reducer')
+        self._fit_step(step='clustering')
```
nsorros, line 79: so you first reduce and then cluster. i wonder whether that removes the ability to generalise to other clustering techniques such as lda, nmf, k means etc. see also here https://scikit-learn.org/stable/auto_examples/applications/plot_topics_extraction_with_nmf_lda.html#sphx-glr-auto-examples-applications-plot-topics-extraction-with-nmf-lda-py. i think k means is propably fine. nmf works but does it work as expected? lda i think only makes sense before reducing

wellcomeml/ml/clustering.py
```diff
@@ -0,0 +1,321 @@
...
+
+            param_grid = {'reducer': {'min_dist': [0.0, 0.2], 'n_neighbors': [
+            2,3,5], 'metric': ['cosine', 'euclidean']},
+            'clustering': {'min_samples': [2, 5], 'eps': [0.5, 1, 1.5]}}
+
+            n_cluster_range (2-uple of ints): A 2-uple describing the
+            max and min number of clusters (e.g.: (10, 20)). If unset,
+            will just choose the best silhopuette
+
+            max_noise (float in [0,1]): The maximum fraction of points
```
nsorros, line 118: is this param only applicable to dbscan?

wellcomeml/ml/clustering.py
```diff
@@ -0,0 +1,321 @@
...
+        if step == 'vectorizer':
+            self.embedded_points = self.vectorizer.fit_transform(X)
+        elif step == 'reducer':
+            self.reduced_points = \
+                self.reducer.fit_transform(self.embedded_points)
+        elif step == 'clustering':
+            self.clustering.fit(self.reduced_points)
+            self.cluster_ids = self.clustering.labels_
+
+    def optimise(self, X, param_grid, n_cluster_range=None, max_noise=0.2,
```
nsorros, line 97: is it possible to try pipeline and gridsearch with the memory param to see whether that is redundant?

nsorros: if it is you can still keep the function but possibly utilise the sklearn functions inside to optimise

wellcomeml/ml/clustering.py
nsorros: that will be a nice addition

wellcomeml/ml/clustering.py
```diff
@@ -0,0 +1,321 @@
...
+                       self.vectorizer.vectorizer.vocabulary_.items()}
+
+        # Initialise a dictionary to calculate the aggregated vector for each
+        # cluster
+
+        aggregated = {key: np.zeros(X_transformed.shape[1])
+                      for key in self.cluster_ids}
+
+        for i in range(len(X_transformed)):
+            aggregated[self.cluster_ids[i]] += X_transformed[i]
```
nsorros, line 249: this is a hack right? you are adding the tfidf values for all the examples of a cluster. actually the addition might make some sense since the denominator for each term i.e. the idf score of the term is the same so it would be a combined count divided by idf for the term approx. one things to note here is that this is a particular problem of dbscan though as kmeans, nmf and lda will come up with a prototype representation for each cluster which translates to some words easily. so maybe it would be good to add an if then else to check for dbscan and add not implemented yet for kmeans, nmf, lda etc

wellcomeml/ml/clustering.py
```diff
@@ -0,0 +1,321 @@
...
+
+        if not used_word:
+            current_kw_list += [new_word]
+
+        i += 1
+
+    return current_kw_list
+
+
+def _clustering_score(estimator, X, y=None):
```
nsorros, line 302: it is worth noting that silhouete is only one of possible scores. coherence has been shown to be a better predictor of topic quality https://radimrehurek.com/gensim/models/coherencemodel.html#gensim.models.coherencemodel.CoherenceModel and https://mimno.infosci.cornell.edu/info6150/readings/N10-1012.pdf. It would be great if we add that now or design the code in such a way that coherence can easily be added in the future

nsorros (commented):
There are some discussion points that are worth ironing before merging
* using additional measures such as coherence instead of silhouete and designing the code to anticipate that
* using additional clustering techniques such as kmeans, nmf and lda
* understanding whether optimise can use standard sklearn functions

nsorros:
To fix the flake8 error once #121 is merged, rebase against master and force push

wellcomeml/ml/clustering.py
aCampello: I agree that LDA only makes sense if you have the raw vectors. For uMAP specifically you often reduce to an intermediate dimension for clustering and a further reduction to 2 for visualisation, although the intermediate reduction did not help my case.

For the time being, I added a possibility of skipping reduction by a parameter `cluster_reduced` in https://github.com/wellcometrust/WellcomeML/pull/109/commits/b0910f00b612a853a9073c3b7d0c09ccb04208b3

wellcomeml/ml/clustering.py
aCampello: Certain algorithms (`DBScan`, `Optics`, `HDBscan`) which are thing are referred to as `noise clustering` do not cluster everything, and assign -1 to some borderline points. For other algorithms, the requirement will be vacuously satisfied, as `noise = 0`.

aCampello:
Hi @nsorros , I have made the proposed changes. Could you re-review the `.optimise` function. It was a _tour de force_ because:

1. Sklearn does not accept DBScan as part of a traditional pipeline because it doesn't have .predict
2. The scoring function had to be re-factored (`_clustering_score`) 
3. The `memory` parameter of Pipelines works fine, however because the silhouette uses the transfomation and you cannot really access the memory outside the scope of grid_search, I had to re-factor wellcomeml's `Vectorizer` to allow it to cache the transformation in itself.

Now I think the main points have been addressed namely: use sklearn for pipelines and make it work with other clustering methods. As for the point about new metrics I could address it in the future.

aCampello:
I'm addressing co-herence here: https://github.com/wellcometrust/WellcomeML/issues/129

PR #111, opened 2020-08-31
Enhance NN implementations with additional parameters

### LINKED ISSUE #112: Add l2 param to bilstm and cnn
so that l2 regularised can be controlled with a param from the user

### LINKED ISSUE #113: Add dropout to embeddings in bilstm and cnn
A common implementation of dropout in embeddings is to drop the entire word which has shown good results in a couple of papers

### LINKED ISSUE #114: Add param to control last dense unit size
At the moment size is 20 but for example when predicting 4k labels a 20 dimension vector is possibly quite small

### LINKED ISSUE #115: Add parameter to control number of attention heads
At the moment the heads default to the sequence length which results in the same output vector so a residual connection can be added. But in long sequences these might be too many heads and additionally residual is not that often used with attention

### LINKED ISSUE #116: Add tensordboard as default callback and the ability to add more callbacks

### LINKED ISSUE #117: Add precision and recall metrics and the ability to pass custom metrics

### LINKED ISSUE #118: Add parameter to control how the vectors from the sequences are combined to create feature layer
for example you can average them , max them , sum them, concat them...

nsorros:
Description
---

- Add param to control l2. Closes #112 
- Add param to control last dense layer size. Closes #114 
- Add param to control number of attention heads. Closes #115 
- Add param to add custom callbacks and add tensorboard as default. Closes #116
- Add param to add custom metrics and add precision, recall as defaults. Closes #117 
- Add param to control how the feature vector is created. Closes #118 
- Add dropout after the embeddings layer. Closes #113 

Checklist
---

- [x] Added tests

### DIFF
--- a/tests/test_bilstm.py
+++ b/tests/test_bilstm.py
@@ -43,3 +43,22 @@ def test_multilabel():
     model.fit(X, Y)
     assert model.score(X, Y) > 0.4
     assert model.predict(X).shape == (5, 4)
+
+def test_attention():
+    X = [
+        "One",
+        "One only",
+        "Two nothing else",
+        "Two and three"
+    ]
+    Y = np.array([0, 0, 1, 1])
+
+    model = Pipeline([
+        ('vec', KerasVectorizer()),
+        ('clf', BiLSTMClassifier(
+                    nb_epochs=10, 
+                    attention=True,
+                    attention_heads=10))
+    ])
+    model.fit(X, Y)
+    assert model.score(X, Y) > 0.6
--- a/tests/test_bilstm.py
+++ b/tests/test_bilstm.py
@@ -44,6 +44,7 @@ def test_multilabel():
     assert model.score(X, Y) > 0.4
     assert model.predict(X).shape == (5, 4)
 
+
 def test_attention():
     X = [
         "One",
@@ -56,7 +57,7 @@ def test_attention():
     model = Pipeline([
         ('vec', KerasVectorizer()),
         ('clf', BiLSTMClassifier(
-                    nb_epochs=10, 
+                    nb_epochs=10,
                     attention=True,
                     attention_heads=10))
     ])
--- a/wellcomeml/ml/bilstm.py
+++ b/wellcomeml/ml/bilstm.py
@@ -6,7 +6,6 @@ from sklearn.metrics import f1_score
 import tensorflow as tf
 
 from wellcomeml.ml.attention import HierarchicalAttention
-from wellcomeml.ml.keras_utils import Metrics
 
 TENSORBOARD_LOG_DIR = "logs/scalars/" + datetime.now().strftime("%Y%m%d-%H%M%S")
 METRIC_DICT = {
@@ -17,6 +16,7 @@ METRIC_DICT = {
 CALLBACK_DICT = {
     'tensorboard': tf.keras.callbacks.TensorBoard(log_dir=TENSORBOARD_LOG_DIR)
 }
+
 
 class BiLSTMClassifier(BaseEstimator, ClassifierMixin):
     def __init__(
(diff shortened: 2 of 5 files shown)


lizgzil (approved):
Couldn't see anything buggy, but don't have familiarity with attention to know if those parts have been implemented correctly.

Since the functions (fit and build_model) within BiLSTMClassifier and CNNClassifier are very similar if not the same, is there a plan to create a general function script where we can call on them in both classes?

nsorros:
> Couldn't see anything buggy, but don't have familiarity with attention to know if those parts have been implemented correctly.
> 
> Since the functions (fit and build_model) within BiLSTMClassifier and CNNClassifier are very similar if not the same, is there a plan to create a general function script where we can call on them in both classes?

I opened an issue this morning on that #119

wellcomeml/ml/bilstm.py
```diff
@@ -1,10 +1,21 @@
...
 from wellcomeml.ml.attention import HierarchicalAttention
-from wellcomeml.ml.keras_utils import Metrics
+
+TENSORBOARD_LOG_DIR = "logs/scalars/" + datetime.now().strftime("%Y%m%d-%H%M%S")
+METRIC_DICT = {
+    'precision': tf.keras.metrics.Precision(name='precision'),
+    'recall': tf.keras.metrics.Recall(name='recall'),
+    'auc': tf.keras.metrics.AUC(name='auc')
+}
+CALLBACK_DICT = {
```
nsorros, line 16: @aCampello and @lizgzil, i have some second thoughts about the name of this variable. is it clearer if i call them dispatcher as @aCampello has done in other places. any names that could avoid using the word dict

tests/test_bilstm.py
```diff
@@ -43,3 +43,23 @@ def test_multilabel():
...
+
+    model = Pipeline([
+        ('vec', KerasVectorizer()),
+        ('clf', BiLSTMClassifier(
+                    nb_epochs=10,
+                    attention=True,
+                    attention_heads=10))
+    ])
+    model.fit(X, Y)
+    assert model.score(X, Y) > 0.6
```
aCampello, line 65: Wondering if this `0.6` is going to annoy our git pipelines in the future? We do `0.4` for other cases.

wellcomeml/ml/bilstm.py
aCampello: I don't have a strong opinion about it. I think dispatch is an old name that comes from other languages, so I often use it but wouldn't say it is best naming practice. It is often when the arguments are functions/callables (I might have abused the notation sometimes).

Other options: `METRICS`, `METRIC_MAP`, `NAME_TO_METRIC` or something..


Perhaps `METRIC_MAP`

wellcomeml/ml/bilstm.py
```diff
@@ -67,27 +88,58 @@ def residual_attention(x1):
...
             nb_outputs, activation=output_activation, kernel_regularizer=l2
         )(x)
-        self.model = tf.keras.Model(inp, out)
+        model = tf.keras.Model(inp, out)
 
         optimizer = tf.keras.optimizers.Adam(lr=self.learning_rate, clipnorm=1.0)
-        metrics = Metrics(validation_data=(X_val, Y_val))
-        self.model.compile(optimizer=optimizer, loss="binary_crossentropy", metrics=[])
+        metrics = [
+            METRIC_DICT[m] if m in METRIC_DICT else m
```
aCampello, line 114: What is the rationale behind this? If the user passes another name (say `f1_score`) it will keep the string?

aCampello (approved):
Looks good to me. Extra freedom of parameter choice are always good.

wellcomeml/ml/bilstm.py
nsorros: The idea is that some metrics can be passed as a string for convenience but others need to be defined explicitly so we support only some metrics via string at the moment and we can expand as we need

PR #126, opened 2020-09-03
Use pipe to make batch predictions in bert classifier

nsorros:
Description
---

Checklist
---

- [ ] Added tests

### DIFF
--- a/wellcomeml/ml/bert_classifier.py
+++ b/wellcomeml/ml/bert_classifier.py
@@ -213,22 +213,22 @@ class BertClassifier(BaseEstimator, ClassifierMixin):
         return self
 
     def predict(self, X):
-        def binarize_output(x):
-            cats = self.nlp(x).cats
+        def binarize_output(doc):
+            cats = doc.cats
             out = [
                 1 if cats[label] > self.threshold else 0 for label in self.unique_labels
             ]
             return out
-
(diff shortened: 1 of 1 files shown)


wellcomeml/ml/bert_classifier.py
```diff
@@ -213,22 +213,22 @@ def partial_fit(self, X, Y, classes=None):
...
-            cats = self.nlp(x).cats
+        def binarize_output(doc):
+            cats = doc.cats
             out = [
                 1 if cats[label] > self.threshold else 0 for label in self.unique_labels
             ]
             return out
-
-        return np.array([binarize_output(x) for x in X])
+        doc_gen = self.nlp.pipe(X)
```
lizgzil, line 222: just looking at Inez's examples, are you sure you dont have to `list()` this?

nsorros: if you want to transform from a generator to a list you can use list but both work in a for loop

nsorros: i named it generator to remind myself that is a genator because when i was testing the speed impact it seems that it was going really fast but it was only because a genetator was running the predictions only when called. anyway details.

lizgzil (approved):
👍

PR #136, opened 2020-09-16
Add early stopping in BiLSTM and CNN

### LINKED ISSUE #139: CNN and BiLSTM do not support early stopping
Both CNN and BiLSTM currently run training for as many epochs as the user has instructed with no ability to stop when the loss stops decreasing.

Tensorflow provides a callback for that which can easily be added and also allows to restore best weights.

nsorros:
Description
---
Fixes #139 

Adds param early_stopping with a default patience of 5 epochs and default ability to restore best weights. In the future these can also be params.

Checklist
---

- [x] Added tests

### DIFF
--- a/wellcomeml/ml/bilstm.py
+++ b/wellcomeml/ml/bilstm.py
@@ -35,7 +35,7 @@ class BiLSTMClassifier(BaseEstimator, ClassifierMixin):
         metrics=["precision", "recall"],
         callbacks=["tensorboard"],
         feature_approach="max",
-        early_stopping = False
+        early_stopping=False
     ):
         self.learning_rate = learning_rate
         self.batch_size = batch_size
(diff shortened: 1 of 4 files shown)


wellcomeml/ml/bilstm.py
```diff
@@ -133,6 +135,10 @@ def fit(self, X, Y, embedding_matrix=None, *_):
             CALLBACK_DICT[c] if c in CALLBACK_DICT else c
             for c in self.callbacks
         ]
+        if self.early_stopping:
+            early_stopping = tf.keras.callbacks.EarlyStopping(
```
aCampello, line 139: That is such a useful callback - I did not know that. Does it just monitor the training loss, or anyway you can make it monitor validation too?

aCampello (approved):

wellcomeml/ml/bilstm.py
nsorros: It monitors validation loss by default but you can change to any of the available metrics https://www.tensorflow.org/api_docs/python/tf/keras/callbacks/EarlyStopping

PR #137, opened 2020-09-16
Add sparse_y param to BiLSTM and CNN that streams sparse Y

### LINKED ISSUE #138: No support for sparse target vectors in BiLSTM and CNN
Our BiLSTM and CNN architecture do not work with a sparse target vector Y. This is relevant especially for MeSH tagging where the Y vector in its dense form is approx 200GB. Even though tensorflow seems to support sparse multiclass classification with sparse categorical cross entropy loss, something similar does not seem to exist for multilabel.

One way around the problem is to stream the data and densify only during the batch used which is what I have been using in my experiments with these architectures.

nsorros:
Description
---
Fixes #138 

Solves the problem of tensorflow not working with sparse Y vectors like the ones we have for MeSH by introducing a param sparse_y which forces the data to be streamed and converts to dense as per needed which is more memory friendly than densifying the entire Y.

Checklist
---

- [x] Added tests

### DIFF
--- a/wellcomeml/ml/bilstm.py
+++ b/wellcomeml/ml/bilstm.py
@@ -1,4 +1,5 @@
 from datetime import datetime
+import math
 
 from sklearn.model_selection import train_test_split
 from sklearn.base import BaseEstimator, ClassifierMixin
@@ -35,7 +36,8 @@ class BiLSTMClassifier(BaseEstimator, ClassifierMixin):
         metrics=["precision", "recall"],
         callbacks=["tensorboard"],
         feature_approach="max",
-        early_stopping=False
+        early_stopping=False,
+        sparse_y = False
     ):
         self.learning_rate = learning_rate
         self.batch_size = batch_size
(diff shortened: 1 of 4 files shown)


wellcomeml/ml/bilstm.py
```diff
@@ -52,6 +56,17 @@ def __init__(
         self.callbacks = callbacks
         self.feature_approach = feature_approach
         self.early_stopping = early_stopping
+        self.sparse_y = sparse_y
+
+    def _yield_data(self, X, Y, batch_size, shuffle=True):
```
aCampello, line 61: Does this mean that you will only use a data generator is sparse? I thought the `if self.sparse_y` could be inside _yield_data instead, before converting to dense.

aCampello (approved):
Looks good, with one minor point.

wellcomeml/ml/bilstm.py
nsorros: I was thinking that i can add that in the future but why not now. Very good point.

nsorros: 11b36edcae3d4ca9ac9fe54da9cba35cbcfce2fd

PR #143, opened 2020-09-28
Add MirroredStrategy in case of multiple GPUs

### LINKED ISSUE #144: Add multi GPU training to NN architectures
Tensorflow supports multi cpu and 1 gpu training by default but not multi gpus. This can easily be added with a Mirrored strategy in the case of multiple CPUs. This could be a default imo as you would only be in a multi gpu instance if you wanted to use more than one GPU.

nsorros:
Description
---

Fixes #144. If multiple GPUs exist in the environment CNN and BiLSTM will use them as default. This extend the default multi CPU and 1 GPU default behaviour which makes use of a GPU or multiple CPUs automatically if present.

This is a change that is hard to test as it might require a multi GPU environment to verify. Let me know if you have any ideas how to test.

Adds
Checklist
---

- [ ] Added tests

### DIFF
--- a/wellcomeml/ml/bilstm.py
+++ b/wellcomeml/ml/bilstm.py
@@ -149,7 +149,7 @@ class BiLSTMClassifier(BaseEstimator, ClassifierMixin):
 
         nb_outputs = Y.max() if not self.multilabel else Y.shape[1]
 
-        if tf.config.list_physical_devices('gpu'):
+        if tf.config.list_physical_devices('GPU'):
             strategy = tf.distribute.MirroredStrategy()
         else:
             strategy = tf.distribute.get_strategy()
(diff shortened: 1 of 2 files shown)


wellcomeml/ml/bilstm.py
```diff
@@ -130,7 +131,11 @@ def residual_attention(x1):
         )(x)
         model = tf.keras.Model(inp, out)
 
-        optimizer = tf.keras.optimizers.Adam(lr=self.learning_rate, clipnorm=1.0)
+        strategy = tf.distribute.get_strategy()
+        if isinstance(strategy, tf.distribute.MirroredStrategy):
+            optimizer = tf.keras.optimizers.Adam(lr=self.learning_rate)
+        else:  # clipnorm is only supported in default strategy
```
aCampello, line 137: That's weird! Clipnorm is super important - do you know why?

aCampello (approved):
I agree that this is hard to test. As long as it doesn't break the existing code, I am happy to accept.

wellcomeml/ml/bilstm.py
nsorros: No idea and could not find online either.

Issue #146, opened 2020-10-01
__all__ should always be a list of strings


aCampello:
In our modules, `__all__` should always be a list of strings, otherwise we get warnings when generating the docs, e.g.:

```
WARNING: __all__ should be a list of strings, not [<class 'wellcomeml.spacy.spacy_doc_to_prodigy.SpacyDocToProdigy'>] (in module wellcomeml.spacy) -- ignoring __all__
```

(and also will not be able to import functions using `from module import *`), according to the docs: https://docs.python.org/3/reference/simple_stmts.html#the-import-statement.

Have to fix

- [x] `ml/__init__.py`
- [x] `datasets/__init__.py`
- [x] `io/__init__.py`
- [x] `spacty/__init__.py`
- [x] `metrics/__init__.py`

PR #147, opened 2020-10-01
Fix __all__ to return strings instead of modules

aCampello:
Description
---

Fixes https://github.com/wellcometrust/WellcomeML/issues/146

If we add modules (instead of string) to `__all__` some times it breaks imports (*) as in:

```bash
TypeError                                 Traceback (most recent call last)
<ipython-input-1-c7541d84f48f> in <module>
----> 1 from wellcomeml.ml import *

~/WellcomeML/build/virtualenv/lib/python3.7/importlib/_bootstrap.py in _handle_fromlist(module, fromlist, import_, recursive)

~/WellcomeML/build/virtualenv/lib/python3.7/importlib/_bootstrap.py in _handle_fromlist(module, fromlist, import_, recursive)

TypeError: Item in wellcomeml.ml.__all__ must be str, not type
```
---

- [ ] ~~Added tests~~ (not necessary)

### DIFF
--- a/wellcomeml/__init__.py
+++ b/wellcomeml/__init__.py
@@ -5,6 +5,6 @@ from .__version__ import (
 from .logger import logger
 
 __all__ = [
-    __name__, __version__, __description__, __url__,
-    __author__, __author_email__, __license__, logger
+    '__name__', '__version__', '__description__', '__url__',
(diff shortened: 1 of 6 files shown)


nsorros (approved):
LGTM 👍

PR #152, opened 2020-10-02
BiLSTM and CNN now accept X, Y to be lists

### LINKED ISSUE #131: CNN and BiLSTM should be able to accept X and Y as lists
Currently the models require passing a numpy array but they should work with python lists as well.

One easy fix is to convert at the start.

Currently the functions that require update are:

- [ ] CNN fit: https://github.com/wellcometrust/WellcomeML/blob/84ea8b9365ec5f3506b1fee598f7fdb3caf933cd/wellcomeml/ml/cnn.py#L170 (which also requires fix to _yield_data)
- [ ] CNN predict:
https://github.com/wellcometrust/WellcomeML/blob/84ea8b9365ec5f3506b1fee598f7fdb3caf933cd/wellcomeml/ml/cnn.py#L212
- [ ] CNN score:
https://github.com/wellcometrust/WellcomeML/blob/84ea8b9365ec5f3506b1fee598f7fdb3caf933cd/wellcomeml/ml/cnn.py#L224

- [ ] BiLSTM fit:
https://github.com/wellcometrust/WellcomeML/blob/84ea8b9365ec5f3506b1fee598f7fdb3caf933cd/wellcomeml/ml/bilstm.py#L141
- [ ] BiLSTM predict:
https://github.com/wellcometrust/WellcomeML/blob/84ea8b9365ec5f3506b1fee598f7fdb3caf933cd/wellcomeml/ml/bilstm.py#L185
- [ ] BiLSTM score:
https://github.com/wellcometrust/WellcomeML/blob/84ea8b9365ec5f3506b1fee598f7fdb3caf933cd/wellcomeml/ml/bilstm.py#L197

Xx-Ashutosh-xX:
Description
---
Have converted X and Y to a NumPy array. If a list might get passed to the function then it will be converted to a NumPy array

Fixes #131 

Checklist
---

- [X] Added link to Github issue or Trello card
- [ ] Added tests

P.S: What do you mean by adding tests?

### DIFF
--- a/wellcomeml/ml/bilstm.py
+++ b/wellcomeml/ml/bilstm.py
@@ -139,8 +139,10 @@ class BiLSTMClassifier(BaseEstimator, ClassifierMixin):
         return model
 
     def fit(self, X, Y, embedding_matrix=None, *_):
-        X = np.array(X)
-        Y = np.array(Y)
+        if isinstance(X, list):
+            X = np.array(X)
+        if isinstance(Y, list):
+            Y = np.array(Y)
         sequence_length = X.shape[1]
         vocab_size = X.max() + 1
 
@@ -185,7 +187,8 @@ class BiLSTMClassifier(BaseEstimator, ClassifierMixin):
         return self
 
     def predict(self, X, *_):
-        X = np.array(X)
+        if isinstance(X, list):
+            X = np.array(X)
         if self.sparse_y:
             Y_pred = []
             for i in range(0, X.shape[0], self.batch_size):
@@ -198,8 +201,10 @@ class BiLSTMClassifier(BaseEstimator, ClassifierMixin):
             return self.model(X).numpy() > 0.5
 
     def score(self, X, Y):
-        X = np.array(X)
-        Y = np.array(Y)
+        if isinstance(X, list):
+            X = np.array(X)
+        if isinstance(Y, list):
+            Y = np.array(Y)
         Y_pred = self.predict(X)
         return f1_score(Y, Y_pred, average="micro")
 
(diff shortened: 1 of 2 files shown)


aCampello:
Thanks a lot for submitting this @Xx-Ashutosh-xX. I don't think we need to add tests.

However your PR has failed, due to the following tests:
```
FAILED tests/test_bilstm.py::test_sparse - IndexError: tuple index out of range
741FAILED tests/test_cnn.py::test_sparse - IndexError: tuple index out of range
```

wellcomeml/ml/bilstm.py
```diff
@@ -139,6 +139,8 @@ def residual_attention(x1):
         return model
 
     def fit(self, X, Y, embedding_matrix=None, *_):
+        X = np.array(X)
```
aCampello, line 142: This will fail when X is an sparse array (and when it is sparse, we don't want to convert it anyway!), so you probably need to check first whether it is a csr_matrix. 

Take a look at `tests/test_cnn.py:test_sparse`

Xx-Ashutosh-xX: I am thinking if check they data type and it turns out to be a list then only we convert it ... ?

aCampello: Yes, that is one possible solution. I will let @nsorros give the final word!

nsorros (commented):
I think the solution looks good. Only issue is that all tests have Y as a numpy array, I would add one test with Y being a list. X seems to always be a list in the tests. Let me know if you need help.

nsorros (approved):
As a lot of time has passed, I think it is time to merge this 👍 We can add a test later.

Issue #35, opened 2020-04-17
Add parameters for word embeddings to KerasVectorizer


nsorros:
#### Description

When `build_embedding_matrix(embeddings_path)` is called - if embeddings_path is not found then download the word embedding given in parameter `word_vectors`. Default is `word_vectors=glove`.

#### Files

https://github.com/wellcometrust/WellcomeML/blob/127c6110732f993f41bfa1123ca594ee0027170e/wellcomeml/ml/keras_vectorizer.py#L24

#### Tasks

- [ ] Add `word_vectors` parameter to `build_embedding_matrix`
- [ ] Add test to https://github.com/wellcometrust/WellcomeML/blob/master/tests/test_keras_vectorizer.py to deal with case where the file isn't found in `build_embedding_matrix`

#### Original issue

Add parameter `word_vectors=glove` that downloads vectors if not available and creates embedding matrix

aCampello:
When we do this, can we add some tests to KerasVectorizer? I think it's one with very small coverage.

nsorros:
👍

lizgzil:
@nsorros I've kept your original issue wording at the bottom of the description - please can you check whether my interpretation is correct

nsorros:
> @nsorros I've kept your original issue wording at the bottom of the description - please can you check whether my interpretation is correct

perfect

nsorros:
@lizgzil i think this is resolved

PR #155, opened 2020-10-12
Keras vectorizer build_embedding_matrix additions

lizgzil:
Description
---

Fixing https://github.com/wellcometrust/WellcomeML/issues/35

I can't actually see any usage of the "build_embedding_matrix" function in other projects, so I'm hoping setting `embeddings_path=None` doesn't break anything.

I've set it so that the function can take a gensim word vector name if the local path isn't given. But can change the logic if this isn't ideal, I wasn't sure whether it'd be better to just give a default model name?

Checklist
---

- [ ] Added tests

### DIFF
--- a/wellcomeml/ml/keras_vectorizer.py
+++ b/wellcomeml/ml/keras_vectorizer.py
@@ -8,8 +8,6 @@ import gensim.downloader as api
 
 from os import path
 
-from wellcomeml.logger import logger
-
 
 class KerasVectorizer(BaseEstimator, TransformerMixin):
     def __init__(self, vocab_size=None, sequence_length=None, oov_token="<OOV>"):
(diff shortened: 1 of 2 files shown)


wellcomeml/ml/keras_vectorizer.py
```diff
@@ -21,20 +26,58 @@ def transform(self, X, *_):
...
+            if path.exists(embeddings_path):
+                embeddings_index = {}
+                with open(embeddings_path) as f:
+                    for line in f:
+                        word, coefs = line.split(maxsplit=1)
+                        coefs = np.fromstring(coefs, "f", sep=" ")
+                        embeddings_index[word] = coefs
+                    emb_dim = len(coefs)
+            else:
+                logger.error("Incorrect local embeddings path")
```
nsorros, line 54: should we log an error or throw an exception. does the pipeline work if you get None back?

wellcomeml/ml/keras_vectorizer.py
```diff
@@ -21,20 +26,58 @@ def transform(self, X, *_):
...
         return pad_sequences(sequences, maxlen=self.sequence_length)
 
-    def build_embedding_matrix(self, embeddings_path):
-        embeddings_index = {}
-        with open(embeddings_path) as f:
-            for line in f:
-                word, coefs = line.split(maxsplit=1)
-                coefs = np.fromstring(coefs, "f", sep=" ")
-                embeddings_index[word] = coefs
+    def build_embedding_matrix(self, embeddings_path=None, word_vectors=None):
```
nsorros, line 29: i wonder whether the two arguments should be one, check the `from_pretrained` argument in transformers https://huggingface.co/transformers/main_classes/model.html#transformers.PreTrainedModel.from_pretrained.

so the idea would be that the user passes either a path to embeddings stored locally or the name of a pretrained embedding. those embeddings are downloaded and cached so second time do not need to download again

nsorros (changes requested):
That is a great addition @lizgzil, there is a discussion point on whether we want to have one argument that received either a path or a name and if it receives a name it downloads and caches the embeddings. I think this would be a nice approach that follows ideas elsewhere in our library as well.

wellcomeml/ml/keras_vectorizer.py
lizgzil: yeh i think youre right it should really throw an exception and stop

lizgzil:
@nsorros I've made those changes. It looks like when you download a gensim model it will cache it automatically and get the file from there going forward - so didn't need to write anything for this. 

"Gensim has a gensim.downloader module for programmatically accessing this data. The module leverages a local cache that ensures data is downloaded at most once." https://radimrehurek.com/gensim/auto_examples/howtos/run_downloader_api.html

wellcomeml/ml/keras_vectorizer.py
```diff
@@ -21,20 +24,59 @@ def transform(self, X, *_):
...
+        else:
+            raise TypeError("No local or GenSim word embeddings given")
 
-        emb_dim = len(coefs)
         num_words = len(self.tokenizer.word_index) + 1
 
         embedding_matrix = np.zeros((num_words, emb_dim))
         for word, i in self.tokenizer.word_index.items():
-            embedding_vector = embeddings_index.get(word)
+            if local_embeddings:
```
nsorros, line 72: is that needed, why not use the try except block for both cases?

nsorros (approved):
LGTM, one possible consideration.

wellcomeml/ml/keras_vectorizer.py
lizgzil: I think it is because the `get` command wont work for the gensim embeddings case

lizgzil: I think it is because the `get` command wont work for the gensim embeddings case

PR #156, opened 2020-10-16
Feature/batch norm to semantic similarity

aCampello:
Description
---

Towards fine tuning linkage models (https://trello.com/c/9ZrYNkko) I added the flexibility to:

1. Turn dropout on and off for Semantic Similarity
2. Add batch norm

I also brought a method (`.initialise_models()`) previously private, to public, because users might want to initialise it for example for plotting.

These changes allow us to go from model 1:

Screenshot 2020-10-16 at 16 36 17

to model 2:

Screenshot 2020-10-16 at 16 02 32

Checklist
---

- [x] Added tests (Re-purposed a test)

### DIFF
--- a/tests/transformers/test_semantic_similarity.py
+++ b/tests/transformers/test_semantic_similarity.py
@@ -43,6 +43,8 @@ def test_semantic_meta_fit():
     classifier = SemanticEquivalenceMetaClassifier(n_numerical_features=2,
                                                    pretrained="scibert",
                                                    batch_size=2,
+                                                   dropout=True,
+                                                   batch_norm=True,
                                                    eval_batch_size=1)
 
     X = [['This sentence has context_1', 'This one also has context_1', 0.1, 0.2],
--- a/tests/transformers/test_semantic_similarity.py
+++ b/tests/transformers/test_semantic_similarity.py
@@ -1,8 +1,21 @@
 # encoding: utf-8
 import pytest
+import random as python_random
+
+import numpy as np
+import tensorflow as tf
 
 from wellcomeml.ml.bert_semantic_equivalence import \
     SemanticEquivalenceClassifier, SemanticEquivalenceMetaClassifier
+
+
+# Random seeds for test reproducibility
+# Based on keras docs
+# https://keras.io/getting_started/faq/#how-can-i-obtain-reproducible-results-using-keras-during-development
+
+np.random.seed(42)
+python_random.seed(42)
+tf.random.set_seed(42)
 
 
 @pytest.mark.transformers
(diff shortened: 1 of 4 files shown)


tests/transformers/test_semantic_similarity.py
```diff
@@ -1,10 +1,23 @@
...
     SemanticEquivalenceClassifier, SemanticEquivalenceMetaClassifier
 
 
+# Random seeds for test reproducibility
+# Based on keras docs
+# https://keras.io/getting_started/faq/#how-can-i-obtain-reproducible-results-using-keras-during-development
+
+np.random.seed(42)
+python_random.seed(42)
+tf.random.set_seed(42)
```
nsorros, line 18: nice

wellcomeml/ml/bert_semantic_equivalence.py
```diff
@@ -240,19 +242,25 @@ class SemanticEquivalenceMetaClassifier(SemanticEquivalenceClassifier):
     and model initialisation. Fit, predict, score remain intact
     """
 
-    def __init__(self, n_numerical_features, dropout_rate=0.1, **kwargs):
+    def __init__(self, n_numerical_features, dropout=False,
+                 batch_norm=False, dropout_rate=0.1, **kwargs):
```
nsorros, line 246: maybe you want to default to dropout 0 to maintain backwards compatibility

nsorros (approved):
LGTM

nsorros:
What is the rational behind adding batch norm?

aCampello:
> What is the rational behind adding batch norm?

When you add new numerical features, those might be out of scale, which forces us to normalise the features before. Batchnorm helps with this.

wellcomeml/ml/bert_semantic_equivalence.py
aCampello: That's what `dropout=False` does

aCampello:
One the tests fails - investigating

aCampello:
Hah - it seems that with batch norm 3 epochs is not enough for the loss to go down...

aCampello:
[Fine-tune models for linkage](https://trello.com/c/9ZrYNkko/1121-fine-tune-models-for-linkage)

wellcomeml/ml/bert_semantic_equivalence.py
nsorros: maybe you do not need two parameters then :P dropout 0 means no dropout is applied

aCampello: You are right. Maybe a merged too quick. In the next PR I will patch this change in.

PR #159, opened 2020-10-19
Fix error in save when attention is on

### LINKED ISSUE #157: Save CNN with attention fails
When trying to save a CNN model with attention enabled this error is thrown
```
return (name.replace(_ESCAPE_CHAR, _ESCAPE_CHAR + _ESCAPE_CHAR)
AttributeError: 'NoneType' object has no attribute 'replace'
```
This is possibly related to the attention layer not having a name for weights https://github.com/tensorflow/tensorflow/issues/36962. Disabling attention resolves the issue so it definitely related somehow to the attention layer.

nsorros:
Description
---

Fixes #157 

Checklist
---

- [x] Added tests

### DIFF
--- a/tests/test_bilstm.py
+++ b/tests/test_bilstm.py
@@ -59,12 +59,12 @@ def test_save_load_attention():
     vec = KerasVectorizer()
     X_vec = vec.fit_transform(X)
 
-    model = BiLSTMClassifier()
+    model = BiLSTMClassifier(attention=True)
     model.fit(X_vec, Y)
 
     with tempfile.TemporaryDirectory() as tmp_dir:
         model.save(tmp_dir)
-        loaded_model = BiLSTMClassifier(attention=True)
+        loaded_model = BiLSTMClassifier()
         loaded_model.load(tmp_dir)
         assert hasattr(loaded_model, 'model')
         assert loaded_model.score(X_vec, Y) > 0.6
--- a/tests/test_bilstm.py
+++ b/tests/test_bilstm.py
@@ -46,6 +46,27 @@ def test_save_load():
         assert hasattr(loaded_model, 'model')
         assert loaded_model.score(X_vec, Y) > 0.6
 
+def test_save_load_attention():
+    X = [
+        "One",
+        "One only",
+        "Two nothing else",
+        "Two and three"
+    ]
+    Y = np.array([0, 0, 1, 1])
+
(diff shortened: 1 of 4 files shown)


tests/test_bilstm.py
```diff
@@ -46,6 +46,27 @@ def test_save_load():
...
+
+    vec = KerasVectorizer()
+    X_vec = vec.fit_transform(X)
+
+    model = BiLSTMClassifier()
+    model.fit(X_vec, Y)
+
+    with tempfile.TemporaryDirectory() as tmp_dir:
+        model.save(tmp_dir)
+        loaded_model = BiLSTMClassifier(attention=True)
```
lizgzil, line 67: should `attention=True` be added to like 61 rather than here? i.e. arent you testing whether you can load a model with attention?

nsorros: oups, good point

nsorros: 5a7a87138b8a20a7bcda9b2518021ef06bf6ace1

lizgzil (approved):
looks good to me

nsorros:
[Fix save in train in grants tagger](https://trello.com/c/AE2bMBhP/1114-fix-save-in-train-in-grants-tagger)

aCampello:
I was able to reproduce the test failure by running the spacy_partial_fit test 1000 times locally. 

That test is not deterministic, but we are fixing the numpy/python seed in the transformers test. Might be worth changing the seed before that test.

aCampello:
The other solution is to `assert model.score(X, Y) > 0` (i.e., the model is learning SOMETHING....)

aCampello (changes requested):
Please see commend below about the tests that fail. They are unrelated to this PR, but I think it is just an unlucky combination of setting the wrong python seed.

nsorros:
> Please see commend below about the tests that fail. They are unrelated to this PR, but I think it is just an unlucky combination of setting the wrong python seed.

If that is the explanation then the tests should be failing when you merged as well

aCampello (approved):
I am happy with the change https://github.com/wellcometrust/WellcomeML/pull/159/commits/6dbe825ad9c607bf0e8721621cc4685425028a34

Even though we didn't get to the bottom of why these two lines interfere with spacy, I don't think it is risky to reduce the test.

PR #166, opened 2020-10-26
Add predict_proba and threshold in cnn,bilstm

nsorros:
Description
---

Add proba function to NN to be able to return probabilities and also add threshold to be able to predict with different thresholds

Checklist
---

- [x] Added tests

### DIFF
--- a/tests/test_bilstm.py
+++ b/tests/test_bilstm.py
@@ -162,3 +162,41 @@ def test_early_stopping():
     # a lot of time to finish running this test
     model.fit(X, Y)
     assert model.score(X, Y) > 0.6
+
+
+def test_predict_proba():
+    X = [
+        "One",
+        "One only",
+        "Two nothing else",
+        "Two and three"
+    ]
+    Y = np.array([0, 0, 1, 1])
+
+    model = Pipeline([
+        ('vec', KerasVectorizer()),
+        ('clf', BiLSTMClassifier())
+    ])
+    model.fit(X, Y)
+    Y_pred_prob = model.predict_proba(X)
+    assert sum(Y_pred_prob >= 0) == Y.shape[0]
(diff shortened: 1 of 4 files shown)


nsorros:
[Add proba function and threshold parameter to NN architectures to be able to lower the threshold for predictions](https://trello.com/c/ZnVdrg6B/1133-add-proba-function-and-threshold-parameter-to-nn-architectures-to-be-able-to-lower-the-threshold-for-predictions)

tests/test_bilstm.py
```diff
@@ -162,3 +162,41 @@ def test_early_stopping():
...
+
+
+def test_predict_proba():
+    X = [
+        "One",
+        "One only",
+        "Two nothing else",
+        "Two and three"
+    ]
+    Y = np.array([0, 0, 1, 1])
```
aCampello, line 174: Can we add a test where Y is a sparse array ? I think the tests are missing that part of the code.

aCampello (approved):
No major comments, just a suggestion about adding a sparse array to the tests so the code coverage doesn't miss it.

tests/test_bilstm.py
nsorros: good point

nsorros:
@aCampello btw i altered the code a bit since sparse_y does not make sense for probs which are dense. let me know if this does not make sense

tests/test_bilstm.py
nsorros: As it turns out there is a test for sparse.
https://github.com/wellcometrust/WellcomeML/blob/6b7503a043157edf5e40d22b9c2c312d255b60ae/tests/test_cnn.py#L96

Did you mean about proba? I removed sparse_y from proba as it did not make sense

aCampello:
Makes sense, actually!

PR #170, opened 2020-11-05
Multi gpu predict

nsorros:
Description
---

This PR adds the ability to use multiple GPUs for predict. Since I implemented this I found out that 1GPU is enough in MeSH for prediction but I do not see a harm in adding that. What it will do is, if there are multiple GPUs it will use them in the same way that the default behaviour for fit us to use as many CPUs and GPUs are available

Checklist
---

- [ ] Added tests

### DIFF
--- a/wellcomeml/ml/bilstm.py
+++ b/wellcomeml/ml/bilstm.py
@@ -218,11 +218,11 @@ class BiLSTMClassifier(BaseEstimator, ClassifierMixin):
             Y_pred = vstack(Y_pred)
             return Y_pred
         else:
-            return self.model.predict(X).numpy() > self.threshold
+            return self.model.predict(X, self.batch_size).numpy() > self.threshold
 
     def predict_proba(self, X):
         # sparse_y not relevant as probs are dense
-        return self.model.predict(X).numpy()
+        return self.model.predict(X, self.batch_size).numpy()
 
     def score(self, X, Y):
         if isinstance(X, list):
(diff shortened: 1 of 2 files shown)


nsorros:
@lizgzil @aCampello thoughts on this?

nsorros:
@lizgzil could you have a look here as well?

lizgzil (commented):
Maybe worth also doing the `len(tf.config.list_physical_devices('GPU')) > 1:` logic as in
https://github.com/wellcometrust/WellcomeML/pull/208/files
Is it worth maybe making these small changes in PR 208 instead of here and closing?

nsorros:
> Maybe worth also doing the `len(tf.config.list_physical_devices('GPU')) > 1:` logic as in
> https://github.com/wellcometrust/WellcomeML/pull/208/files
> Is it worth maybe making these small changes in PR 208 instead of here and closing?

Either way. I am happy to do in the other PR for consistency

aCampello (approved):
Looks good. Wondering whether there is a way to abstract this away, so we can add in all classes at once (including semantic Bert perhaps)

nsorros:
> Looks good. Wondering whether there is a way to abstract this away, so we can add in all classes at once (including semantic Bert perhaps)

I think this would require a common abstract class for our models, maybe our tensorflow models that implements some basic / common functions. I think it is out of scope for this PR.

PR #171, opened 2020-11-05
new: Add f1 metric and loss functions and GlobalF1 callback (fixes #172)

### LINKED ISSUE #172: Calculate loss based on f1
Something I think we have all been experiencing (especially when dealing with imbalanced classes) is that we evaluate our models based on f1 score, but we optimise using a loss like `binary_crossentropy`.

It would be great if we could evaluate and optimise for the same metric.

An example of how to create an f1 based loss is available here: https://www.kaggle.com/rejpalcz/best-loss-function-for-f1-score-metric#Optimal-loss-function---macro-F1-score

ivyleavedtoadflax:
## Description
Based on #172:

Adds two new functions:
* f1_metric: A metric for passing into keras `model.compile` which will return the batchwise macro f1 score
* f1_loss: A loss function which can be passed to `model.compile` which will optimise for the batchwise macro f1 score.

Expands the Metrics callback:
* Enables the global f1 history (against a validation set) to be saved for reporting purposes.

## Checklist

- [x] Added tests

### DIFF
--- a/tests/metrics/test_f1.py
+++ b/tests/metrics/test_f1.py
@@ -1,13 +1,10 @@
 #!/usr/bin/env python3
 # coding: utf-8
 
-import os
-import tempfile
-
 import numpy as np
-
 import pytest
 import tensorflow as tf
+
 from wellcomeml.metrics import f1_loss, f1_metric
 
 
@@ -85,7 +82,9 @@ def test_f1_metric(data, model):
     """ Test whether the f1_metrics are output"""
 
     model.compile(
-        loss="binary_crossentropy", optimizer="adam", metrics=[f1_metric],
+        loss="binary_crossentropy",
+        optimizer="adam",
+        metrics=[f1_metric],
     )
 
     history = model.fit(
@@ -106,10 +105,12 @@ def test_f1_loss(data, model):
     """ Test to see if it runs, don't test the loss itself """
 
     model.compile(
-        loss=f1_loss, optimizer="adam", metrics=["accuracy"],
+        loss=f1_loss,
+        optimizer="adam",
+        metrics=["accuracy"],
     )
 
-    history = model.fit(
+    model.fit(
         data["X_train"],
         data["y_train"],
         epochs=5,
@@ -117,5 +118,3 @@ def test_f1_loss(data, model):
         batch_size=1024,
         verbose=0,
     )
-
-
--- a/tests/metrics/test_f1.py
+++ b/tests/metrics/test_f1.py
@@ -8,7 +8,7 @@ import numpy as np
 
 import pytest
 import tensorflow as tf
-from wellcomeml.metrics import GlobalF1, f1_loss, f1_metric
+from wellcomeml.metrics import f1_loss, f1_metric
 
 
 @pytest.fixture(scope="module")
@@ -119,26 +119,3 @@ def test_f1_loss(data, model):
     )
 
 
-def test_globalf1_callback(data, model, tmpdir):
-
-    history_path = os.path.join(tmpdir, "test_f1.csv")
-
-    model.compile(
-        loss=f1_loss, optimizer="adam", metrics=["accuracy"],
-    )
-
-    metrics = GlobalF1(
-        validation=(data["X_test"], data["y_test"]), history_path=history_path
-    )
-
-    history = model.fit(
-        data["X_train"],
-        data["y_train"],
-        epochs=5,
-        validation_data=(data["X_test"], data["y_test"]),
-        batch_size=1024,
-        verbose=0,
-        callbacks=[metrics],
-    )
-
-    assert os.path.exists(history_path)
--- a/tests/metrics/test_f1.py
+++ b/tests/metrics/test_f1.py
@@ -1,0 +1,144 @@
+#!/usr/bin/env python3
+# coding: utf-8
+
+import os
+import tempfile
+
+import numpy as np
+
+import pytest
+import tensorflow as tf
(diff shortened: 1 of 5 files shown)


wellcomeml/metrics/f1.py
```diff
@@ -0,0 +1,113 @@
...
+    >>> global_f1 = GlobalF1(validation=(X_test, y_test, history.csv))
+    >>>
+    >>> history = model.fit(
+    >>>     X_train,
+    >>>     y_train,
+    >>>     epochs=100,
+    >>>     validation_data=(X_test, y_test),
+    >>>     batch_size=1024,
+    >>>     verbose=2,
+    >>>     callbacks=[early_stopping, glbal_f1],
```
aCampello, line 71: Typo

wellcomeml/metrics/f1.py
```diff
@@ -0,0 +1,113 @@
...
+    p = tp / (tp + fp + K.epsilon())
+    r = tp / (tp + fn + K.epsilon())
+    f1 = 2 * p * r / (p + r + K.epsilon())
+    f1 = tf.where(tf.math.is_nan(f1), tf.zeros_like(f1), f1)
+
+    return 1 - K.mean(f1)
+
+
+class GlobalF1(Callback):
+    """Calculate global F1, precision, and recall, against a complete test set
```
aCampello, line 59: I think we might already have implemented this. The implementation is different but check out https://github.com/wellcometrust/WellcomeML/blob/master/wellcomeml/ml/keras_utils.py. Maybe your implementation is better, so you could change the one we already have instead of a new class?

aCampello (changes requested):
Thanks a lot for the PR, Matt. I'm really keen to use the f1 loss and see whether this will improve performance to some of my existing projects!

One thing is that you ticked the "added tests" box, but I can't see them. Would be good to have tests to the new metric so we don't decrease coverage too much.

wellcomeml/metrics/f1.py
ivyleavedtoadflax: ooh good spot!

ivyleavedtoadflax:
> Thanks a lot for the PR, Matt. I'm really keen to use the f1 loss and see whether this will improve performance to some of my existing projects!
> 
> One thing is that you ticked the "added tests" box, but I can't see them. Would be good to have tests to the new metric so we don't decrease coverage too much.

oops yep will add them. I wondered why the test coverage was going down, not up!

wellcomeml/metrics/f1.py
ivyleavedtoadflax: https://github.com/wellcometrust/WellcomeML/pull/171/commits/0637fe5bbfe094dbde91b56cc833d977a61fc08c

ivyleavedtoadflax:
This should be ready to go now - it even bumped up overall test coverage by a tad!

tests/metrics/test_f1.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+@pytest.fixture(scope="module")
+def tmpdir(tmpdir_factory):
+    return tmpdir_factory.mktemp("test_f1")
+
+
+@pytest.fixture(scope="module")
+def data():
+    X_train = np.random.random((100, 10))
```
aCampello, line 18: This is a great test fixture. I would say it's not reproducible because of a random seed, but given the type of assertion below, I think it is fine.

wellcomeml/ml/keras_utils.py
```diff
@@ -1,12 +1,34 @@
 from collections import defaultdict
 
-from sklearn.metrics import f1_score, precision_score, recall_score
+import pandas as pd
 import tensorflow as tf
+from sklearn.metrics import f1_score, precision_score, recall_score
 
 
 class Metrics(tf.keras.callbacks.Callback):
```
aCampello, line 8: This is nice. I am impressed that it worked with all the other classes that used it seamlessly

tests/metrics/test_f1.py
ivyleavedtoadflax: Yes that's true, i think as you say, we are probably ok with the simple assertion I make for now. Perhaps there will need to be a time when we make a more concrete assertion and we could adjust the fixture.

wellcomeml/ml/keras_utils.py
```diff
@@ -16,9 +38,26 @@ def on_epoch_end(self, epoch, logs):
...
+    def on_train_end(self, logs={}):
+        """Write metrics to csv file"""
+
+        if self.history_path:
+            history = {}
+            history["f1"] = self.f1s
+            history["precision"] = self.precisions
+            history["recall"] = self.recalls
+            history_df = pd.DataFrame(history)
+            history_df.to_csv(self.history_path, index_label="epoch")
```
aCampello: Pandas seems a bit of a cannon to kill a fly here. Can we save in jsonL?

aCampello (approved):
LGTM. Just wrote down some very minor comments, no need to address them if you don't want. :shipit:

aCampello:
Your flake8 seems to have failed?

wellcomeml/ml/keras_utils.py
ivyleavedtoadflax: Yeah this is true, and I had been thinking the same. Ideally we do want it in csv so that it can be easily read by dvc (the jsonl structure  required for this is a bit convoluted), let me switch this out and use the csv module instead.

wellcomeml/ml/keras_utils.py
ivyleavedtoadflax: haha yes... :crossed_fingers:

PR #178, opened 2020-11-19
Fix scibert download

### LINKED ISSUE #177: BertClassifier scibert path is not working
Originally the way we downloaded SciBert was manually and we were ensuring it is available in models 
https://github.com/wellcometrust/WellcomeML/blob/3c9bd5723043a561967e348448787a6d881c020d/wellcomeml/ml/bert_classifier.py#L79

Since then we have introduced a cache and a checks and a download function which can be leveraged to do the download for you.

nsorros:
Description
---

Fixes #177 

Checklist
---

- [x] Added tests


lizgzil (approved):
`pytest tests/test_bert_classifier.py` pass 👍

PR #182, opened 2020-11-25
Fix BertVectorizer for long sequences and add tests

### LINKED ISSUE #173: Bert vectorizer count of max tokens is wrong
https://github.com/wellcometrust/WellcomeML/blob/16754b0a9806b29b942a4ac67ba9b67253ce97fc/wellcomeml/ml/bert_vectorizer.py#L43

This lines tries to ensure that a max of 512 tokens are passed to BERT but tokens are not number of characters but number of words or subwords to be exact. The most accurate way to count them would be to tokenize using the bert tokenizer but we could also approximate by `x.split()`

nsorros:
Description
---

Fixes #173 

The fix is slightly different than the one suggested in the issue. Let me know what you think.

Checklist
---

- [x] Added tests

### DIFF
--- a/wellcomeml/ml/bert_vectorizer.py
+++ b/wellcomeml/ml/bert_vectorizer.py
@@ -40,14 +40,15 @@ class BertVectorizer(BaseEstimator, TransformerMixin):
         return np.load(path)
 
     def bert_embedding(self, x):
-        # Max sequence length is 512 for BERT
-        if len(x) > 512:
-            embedded_a = self.bert_embedding(x[:512])
-            embedded_b = self.bert_embedding(x[512:])
+        tokenized_x = self.tokenizer.tokenize(x)
+
+        # Max sequence length is 512 for BERT. 510 without CLS and SEP
+        if len(tokenized_x) > 510:
+            embedded_a = self.bert_embedding(" ".join(tokenized_x[:510]))
+            embedded_b = self.bert_embedding(" ".join(tokenized_x[510:]))
             return embedded_a + embedded_b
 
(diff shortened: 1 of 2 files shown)


nsorros:
Interestingly the test that fails succeeds when i run it, need to investigate this a bit

nsorros:
@aCampello @lizgzil

nsorros:
@lizgzil @aCampello this is a relatively small PR

wellcomeml/ml/bert_vectorizer.py
```diff
@@ -40,14 +40,15 @@ def load_transformed(cls, path):
...
     def bert_embedding(self, x):
-        # Max sequence length is 512 for BERT
-        if len(x) > 512:
-            embedded_a = self.bert_embedding(x[:512])
-            embedded_b = self.bert_embedding(x[512:])
+        tokenized_x = self.tokenizer.tokenize(x)
+
+        # Max sequence length is 512 for BERT. 510 without CLS and SEP
+        if len(tokenized_x) > 510:
+            embedded_a = self.bert_embedding(" ".join(tokenized_x[:510]))
```
aCampello, line 47: I see, we are approximating by joining them with " ". It might miss some edge cases, but probably ok!

aCampello (approved):

aCampello:
LGTM

wellcomeml/ml/bert_vectorizer.py
nsorros: whatever the reason for splitting them before, space should make them split again is my theory so its only safe to add spaces there

PR #183, opened 2020-11-30
Upgrade spacy to v2.3

nsorros:
Description
---
This PR updates spacy to v2.3, transformers to <2.9 and spacy-transformers (which dictates how much we can push the previous packages forward) to 0.6.1

Upgrading to v2.3 caused the knowledge base to stop being imported. I fixed that by following the code in the updated link on the top of the script https://spacy.io/usage/training#entity-linker. What was failing was the import `from bin.wiki_entity_linking.train_descriptions import EntityEncoder` which was removed according to https://github.com/explosion/spaCy/issues/6362. I noticed in the code that EntityEncoder was creating the embeddings which are not simply created by `nlp(desc).vector` so i updated the process but i am not 100% sure it works as there are no tests.

Spacy v2.3 release page https://github.com/explosion/spaCy/releases/tag/v2.3.0

Checklist
---

- [ ] Added tests


lizgzil:
@nsorros tests added for the spacy knowledge base, and a little correction. I wanted to add tests for the `spacy_entity_linker.py` too, but there is an issue with this!

Basically when you try to train the entity linker (after training a knowledge base) you get the error:

```
  File "/Users/gallaghe/Code/WellcomeML/wellcomeml/ml/spacy_entity_linking.py", line 77, in train
    texts, annotations, drop=0.2, losses=losses, sgd=optimizer,
  File "/Users/gallaghe/Code/WellcomeML/build/virtualenv/lib/python3.7/site-packages/spacy/language.py", line 526, in update
    proc.update(docs, golds, sgd=get_grads, losses=losses, **kwargs)
  File "pipes.pyx", line 1227, in spacy.pipeline.pipes.EntityLinker.update
  File "doc.pyx", line 626, in sents
ValueError: [E030] Sentence boundaries unset. You can add the 'sentencizer' component to the pipeline with: nlp.add_pipe(nlp.create_pipe('sentencizer')) Alternatively, add the dependency parser, or set sentence boundaries by setting doc[i].is_sent_start.
```

There is a lot online about this error. I found the simple `nlp.add_pipe(nlp.create_pipe('sentencizer'))` addition fix didn't solve this (same error! even though `nlp.pipe_names` contains 'sentencizer'). I also tried adding a 'parser' pipe. Very confusing :/

This didn't come up before, so I'm wondering whether its a 2.3 problem?

nsorros:
> @nsorros tests added for the spacy knowledge base, and a little correction. I wanted to add tests for the `spacy_entity_linker.py` too, but there is an issue with this!
> 
> Basically when you try to train the entity linker (after training a knowledge base) you get the error:
> 
> ```
>   File "/Users/gallaghe/Code/WellcomeML/wellcomeml/ml/spacy_entity_linking.py", line 77, in train
>     texts, annotations, drop=0.2, losses=losses, sgd=optimizer,
>   File "/Users/gallaghe/Code/WellcomeML/build/virtualenv/lib/python3.7/site-packages/spacy/language.py", line 526, in update
>     proc.update(docs, golds, sgd=get_grads, losses=losses, **kwargs)
>   File "pipes.pyx", line 1227, in spacy.pipeline.pipes.EntityLinker.update
>   File "doc.pyx", line 626, in sents
> ValueError: [E030] Sentence boundaries unset. You can add the 'sentencizer' component to the pipeline with: nlp.add_pipe(nlp.create_pipe('sentencizer')) Alternatively, add the dependency parser, or set sentence boundaries by setting doc[i].is_sent_start.
> ```
> 
> There is a lot online about this error. I found the simple `nlp.add_pipe(nlp.create_pipe('sentencizer'))` addition fix didn't solve this (same error! even though `nlp.pipe_names` contains 'sentencizer'). I also tried adding a 'parser' pipe. Very confusing :/
> 
> This didn't come up before, so I'm wondering whether its a 2.3 problem, and maybe in v3 it is fixed.

Unlikely that this will be fixed in v3. I think you should add tests, otherwise the code will just not work if someone is trying to use it

nsorros:
Incorporated to #184

tests/transformers/test_semantic_similarity.py
```diff
@@ -64,7 +64,7 @@ def test_semantic_meta_fit():
 
     y = [1, 1, 0]
 
-    classifier.fit(X, y, epochs=3)
+    classifier.fit(X, y, epochs=5)
```
lizgzil, line 67: how come you did this out of interest?

wellcomeml/ml/bert_semantic_equivalence.py
```diff
@@ -142,7 +142,8 @@ def _tokenize(self, X):
 
     def _prep_data_for_prediction(self, X):
         X_tokenized = self._tokenize(X)
-        return tf.convert_to_tensor(self.model.predict(X_tokenized))
+        predictions = self.model.predict(X_tokenized)[0]  # Issue
```
lizgzil, line 145: fyi what I did to line 238 in predict `self.score(X).argmax(axis=2).flatten()` means that taking the 0th element will be just one prediction now, not the array on predictions for all of X. The flatten() should have sorted the output of predict out so it is indeed just "An array of 0s and 1s" like the prediction function docstring describes.

nsorros:
Just to summarise where we are with this PR.

Upgrading spacy to v2.3 allows us to upgrade spacy-transformers to 0.6 and transformers to <2.9 which is what we have done. Spacy transformers 0.6 works with at most spacy v2.3 and with transformers <2.9 so this is the bottleneck to upgrading those futher. There is a newer version of spacy-transformers 1.0 which works with spacy v3.0 and transformers >3.0 and <4.0 which is addressed in a separate PR #184 

Upgrading transformers to <2.9 allows us to integrate the transformers tests and remove the need for a separate installation of transformers for SemanticEquivalenceClassifier to work. This creates a separate problem though because the tensorflow version that SemanticEquivalenceClassifier was developed on was 2.1.2 which if pinned now causes an error to cnn and bilstm in their ability to save the model. The rest of the WellcomeML library is compatible with the latest tensorflow 2.3.0 (2.4.0?) but SemanticEquivalenceClassifier throws some documented errors #187 and #186 which are related with the default callbacks for tensorboard and metricminibatch_history which are removed from this PR to make the tests pass.

tests/transformers/test_semantic_similarity.py
nsorros: the test was failing with loss increasing for some reason

aCampello (approved):

aCampello:
A couple of issues arise from this merge, such as https://github.com/wellcometrust/WellcomeML/issues/189, https://github.com/wellcometrust/WellcomeML/issues/187, https://github.com/wellcometrust/WellcomeML/issues/186, but we decided that it is probably best to merge it and leave those as issues to be fixed later, so we can move on with the attempt to upgrade to 3.0

nsorros:
Huge 🎉

PR #191, opened 2021-01-21
Rewrite BertClassifier using transformers

### LINKED ISSUE #33: Add native HuggingFace Scibert support

nsorros:
Description
---
This PR rewrites BertClassifier using the hugging face transformers library moving away from spacy-transformers. There are a couple of ways we are doing this

- It gives us more flexibility on the features we can use from transformers as spacy-transformers is a wrapper on top that somehow limits what we can do, for example our ability to use all models available.
- Makes it easier to move towards training our own BERT models since a lot of the components can be reused to create a BERT for language modelling
- It decouples slightly from spacy which is generally bad in being backwards compatible, for example upgrading to spacy v3 caused models trained with bert classifier to break even after making the changes necessary to transition

Fixes #33 

Checklist
---

- [x] Added tests

### DIFF
--- a/wellcomeml/ml/bert_classifier.py
+++ b/wellcomeml/ml/bert_classifier.py
@@ -33,7 +33,7 @@ class BertClassifier(BaseEstimator, TransformerMixin):
     def __init__(self, learning_rate=5e-5, epochs=5, batch_size=8,
                  pretrained="bert-base-uncased", threshold=0.5,
                  validation_split=0.1, max_length=512, multilabel=True,
-                 from_pt=False):
+                 from_pt=False, random_seed=42):
         """
         Args:
            learning_rate(float): learning rate to optimize model, default 5e-5
@@ -55,7 +55,7 @@ class BertClassifier(BaseEstimator, TransformerMixin):
         self.validation_split = validation_split
         self.max_length = max_length
         self.multilabel = multilabel
-        self.random_seed = 42
+        self.random_seed = random_seed
         self.from_pt = from_pt
 
     def _init_model(self, num_labels=2):
--- a/wellcomeml/ml/bert_classifier.py
+++ b/wellcomeml/ml/bert_classifier.py
@@ -1,236 +1,191 @@
-#!/usr/bin/env python
-# coding: utf8
 """
-Train a convolutional neural network for multilabel classification
-of grants
-Adapted from https://github.com/explosion/spaCy/blob/master/examples/training/train_textcat.py
+Implements BertClassifier, an sklearn compatible BERT
+classifier class that can be used for text classification
+tasks
 """
-from spacy_transformers import (
-    TransformersLanguage,
-    TransformersWordPiecer,
-    TransformersTok2Vec,
-)
-from spacy_transformers.model_registry import (
-    register_model,
-    get_last_hidden,
-    flatten_add_lengths,
-)
-from spacy.util import minibatch, compounding
-from spacy._ml import zero_init, logistic
-from thinc.t2v import Pooling, mean_pool
-from thinc.v2v import Affine
-from thinc.api import chain
-from sklearn.metrics import precision_recall_fscore_support, f1_score
-from sklearn.base import BaseEstimator, ClassifierMixin
+from transformers import BertTokenizer, TFBertForSequenceClassification
+from sklearn.base import BaseEstimator, TransformerMixin
 from sklearn.model_selection import train_test_split
-import numpy as np
-import spacy
-import torch
+from sklearn.metrics import f1_score
+import tensorflow as tf
 
-import random
-import time
-
-from wellcomeml.utils import check_cache_and_download
-from wellcomeml.logger import logger
+import math
+import os
 
 
-@register_model("sigmoid_last_hidden")
-def sigmoid_last_hidden(nr_class, *, exclusive_classes=False, **cfg):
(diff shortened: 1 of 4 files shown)


wellcomeml/ml/bert_classifier.py
```diff
@@ -1,236 +1,191 @@
...
-            config={"exclusive_classes": False, "architecture": "sigmoid_last_hidden"},
-        )
-
-        self.nlp.add_pipe(self.textcat, last=True)
+            return input_data
 
-        for label in self.unique_labels:
-            self.textcat.add_label(label)
+    def _prepare_data(self, X, Y=None, shuffle=False, repeat=False):
+        X_vec = self.tokenizer.batch_encode_plus(
```
aCampello, line 87: If I am not mistaken, this was the bottleneck for my processing. If using in a CPU, it doesn't parallelise. I am not sure if it parallelise with multi-GPU either. Wonder if the new version of transformers/torch has fixed this.

wellcomeml/ml/bert_classifier.py
```diff
@@ -1,236 +1,191 @@
...
+            dataset = tf.data.Dataset.from_tensor_slices(
+                (X_vec["input_ids"], X_vec["attention_mask"], Y)
+            )
+        else:
+            dataset = tf.data.Dataset.from_tensor_slices(
+                (X_vec["input_ids"], X_vec["attention_mask"])
+            )
+        dataset = dataset.map(self._transform_data)
+        if shuffle:
+            dataset = dataset.shuffle(100, seed=self.random_seed)
```
aCampello, line 102: Not sure if I understand the variable `buffer_size`, but why do you set to a hard number, 100?

aCampello: To be honest, I also do something weird with `buffer_size` :

https://github.com/wellcometrust/WellcomeML/blob/362635afe40779018183aa535b92f5db7d673624/wellcomeml/ml/bert_semantic_equivalence.py#L178

wellcomeml/ml/bert_classifier.py
```diff
@@ -1,236 +1,191 @@
...
+            train_data, validation_data=val_data, epochs=self.epochs,
+            steps_per_epoch=steps_per_epoch, validation_steps=val_steps_per_epoch)
+        self.losses = history.history["loss"]
         return self
 
+    def _predict_proba(self, X):
+        dataset = self._prepare_data(X)
+        out = self.model.predict(dataset)[0]
+        if self.multilabel:
+            return tf.nn.sigmoid(out).numpy()
```
aCampello, line 148: I see. I do ```tf.keras.activations.sigmoid/softmax```, but I think the functions just call one another on TensorFlow.

aCampello (approved):
I think the classifier looks good. It looks similar to `bert_semantic_equivalence.py`, with the exception of the data transformation. But even there, I think we just use different verbs, but the data structure is the same. 

It would be good to understand the `buffer_size` parameter of the data transformation function too. After this PR is merged, I will probably improve semantic similarity with the fixed batch size and that.

aCampello (approved):
Some very minor comments about random seed and shuffle, but other than that looks good.

wellcomeml/ml/bert_classifier.py
nsorros: Good point, actually my next task is to investigate a bit the tokenisation in order to build a class that we can create a custom tokenizer. So maybe this will give us a way

wellcomeml/ml/bert_classifier.py
nsorros: yeah i noticed. instead of loading all elements into memory, it takes batches of 100 and shuffles

wellcomeml/ml/bert_classifier.py
```diff
@@ -1,236 +1,191 @@
...
-            path = check_cache_and_download("scibert_scivocab_uncased")
-
-            nlp = TransformersLanguage(trf_name=name, meta={"lang": "en"})
-            nlp.add_pipe(nlp.create_pipe("sentencizer"))
-            nlp.add_pipe(TransformersWordPiecer.from_pretrained(nlp.vocab, path))
-            nlp.add_pipe(TransformersTok2Vec.from_pretrained(nlp.vocab, path))
-            self.nlp = nlp
+        self.max_length = max_length
+        self.multilabel = multilabel
+        self.random_seed = 42
```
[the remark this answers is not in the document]
nsorros: 👍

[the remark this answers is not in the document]
nsorros: 0c2dccaac45bdec2cc49c465a41f03002f372a1c

wellcomeml/ml/bert_classifier.py
nsorros: agree

nsorros:
[Rewrite BertClassifier using transformers and tensorflow](https://trello.com/c/E2wlr2l2/1161-rewrite-bertclassifier-using-transformers-and-tensorflow)

PR #204, opened 2021-02-09
Efficient CNN

nsorros:
Description
---

This PR reduces the memory requirement needed for CNN by introducing the ability to pass a tensorflow dataset as an input. In order to unify how data are handled within CNN, when data passed is a list of numpy array, they are converted to a tensorflow dataset as well.

What makes tensorflow dataset light in memory is their ability to initialise from a generator. Actually passing a generator is another way CNN can become memory light but then functionality around batching, shuffling etc would need to be implemented when they come for free with a tensorflow dataset.

Checklist
---

- [x] Added tests

### DIFF
--- a/tests/test_cnn.py
+++ b/tests/test_cnn.py
@@ -18,7 +18,7 @@ def test_vanilla():
 
     model = Pipeline([
         ('vec', KerasVectorizer()),
-        ('clf', CNNClassifier())
+        ('clf', CNNClassifier(batch_size=2))
     ])
     model.fit(X, Y)
     assert model.score(X, Y) > 0.6
@@ -36,7 +36,7 @@ def test_save_load():
     vec = KerasVectorizer()
     X_vec = vec.fit_transform(X)
 
-    model = CNNClassifier()
+    model = CNNClassifier(batch_size=2)
     model.fit(X_vec, Y)
 
     with tempfile.TemporaryDirectory() as tmp_dir:
@@ -59,7 +59,10 @@ def test_save_load_attention():
     vec = KerasVectorizer()
     X_vec = vec.fit_transform(X)
 
-    model = CNNClassifier(attention=True)
+    model = CNNClassifier(
+        batch_size=2,
+        attention=True
+    )
     model.fit(X_vec, Y)
 
     with tempfile.TemporaryDirectory() as tmp_dir:
@@ -87,7 +90,10 @@ def test_multilabel():
     ])
     model = Pipeline([
         ('vec', KerasVectorizer()),
-        ('clf', CNNClassifier(multilabel=True))
+        ('clf', CNNClassifier(
+            batch_size=2,
+            multilabel=True
+        ))
     ])
     model.fit(X, Y)
     assert model.score(X, Y) > 0.4
@@ -133,6 +139,7 @@ def test_attention():
     model = Pipeline([
         ('vec', KerasVectorizer()),
         ('clf', CNNClassifier(
+                    batch_size=2,
                     attention=True,
                     attention_heads=10))
     ])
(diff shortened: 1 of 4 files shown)


nsorros:
[Refactor CNN and train to be memory efficient](https://trello.com/c/pDkyOynd/1175-refactor-cnn-and-train-to-be-memory-efficient)

tests/test_cnn.py
```diff
@@ -151,12 +159,14 @@ def test_early_stopping():
...
         ('vec', KerasVectorizer()),
         ('clf', CNNClassifier(
+                    batch_size=2,
                     early_stopping=True,
                     nb_epochs=10000
         ))
     ])
     # if early_stopping is not working it will take
     # a lot of time to finish running this test
+    # it will also consume the 4MB of logs in travis
```
aCampello, line 169: Can we define that env variable to silence the logs at least in travis (https://github.com/wellcometrust/WellcomeML/issues/209) ?

wellcomeml/ml/cnn.py
```diff
@@ -98,9 +112,53 @@ def _get_distributed_strategy(self):
             strategy = tf.distribute.get_strategy()
         return strategy
 
+    def _init_from_data(self, X, Y):
+        logger.info(
+            "Initializing sequence_length, vocab_size \
+            and nb_outputs from data. This might take a while."
+        )
+        if type(X) is np.ndarray:
```
aCampello, line 120: `isinstance(X, np.ndarray)` will do too

wellcomeml/ml/cnn.py
```diff
@@ -98,9 +112,53 @@ def _get_distributed_strategy(self):
...
+            X = X.batch(self.batch_size)
+
+            # init from iterating over dataset
+            X_max = 0
+            Y_max = 0
+            X_shape = None
+            Y_shape = None
+            steps_per_epoch = 0
+            for X_batch, Y_batch in X:
+                batch_X_max = X_batch.numpy().max()
```
aCampello, line 136: I don't understand this `.numpy().max()` operation..

aCampello: Does this give you the length? I thought that's what you needed to calculate

aCampello (approved):
Took me a while to understand the logic between `.take` and `.skip`, but all in all it's great work! 👍 If you feel like, could put one of the tests into `/example`.

tests/test_cnn.py
nsorros: Good point will do that

wellcomeml/ml/cnn.py
nsorros: possibly better and more consistent, will dot the change

wellcomeml/ml/cnn.py
nsorros: it gives me the maximum integer in X batch. I want to find the vocabulary size.

the length of the columns is also used to define sequence length but the variable is called X_shape.

happy to rename if anything is unclear

wellcomeml/ml/cnn.py
```diff
@@ -180,37 +240,48 @@ def residual_attention(x1):
...
+            steps_per_epoch = self._init_from_data(X, Y)
+
+        if type(X) is np.ndarray:
+            data = self._prepare_data(X, Y, shuffle=True)
+        else:  # tensorflow dataset
+            data = X.batch(self.batch_size)
+
+        train_steps_per_epoch = int((1-self.validation_split) * steps_per_epoch)
+        if train_steps_per_epoch == 0:
+            logger.warning(
```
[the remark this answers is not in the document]
nsorros: it still works but there are no validation metrics. actually early stopping does not work so maybe I should mention

wellcomeml/ml/cnn.py
nsorros: b9bf570

wellcomeml/ml/cnn.py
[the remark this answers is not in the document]
nsorros: 5ae533c

PR #205, opened 2021-02-10
Fix version to latest version on PyPi

### LINKED ISSUE #206: Incosistent version number in master and PyPi
Our latest release `2021.02.0` did not update the internal version which might create problem as pip might complain in the future if someone asks for a specific version in which it will use the version file to check. We need to update to `2020.02.1` asap to fix this as we cannot overwrite `2020.02.0`

nsorros:
Description
---

We forgot to update the version internal during the last release. Unfortunately we cannot overwrite the version in PyPi so this just fixes the problem for those cloning and using from Github. Ideally we want to update to `2021.02.1` and upload to PyPi which I am happy to do.

Fixes #206
 
Checklist
---

- [ ] Added tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# 2021.2.1
+
+Bug fixes:
+- Update version to 2021.2.1 which was out of sync in latest release #206
+
 # 2021.2.0
 
 Major changes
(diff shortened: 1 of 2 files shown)


lizgzil (approved):
good spot

nsorros:
Should we merge this or increase to `2020.02.1` and release? @aCampello @lizgzil I think the second is possibly better

aCampello (approved):
Go for it

aCampello:
I think we need to do both - merge and update.

nsorros:
Released https://pypi.org/project/wellcomeml/2021.2.1/

PR #208, opened 2021-02-10
Enable multi gpu setup when more than one GPUs

### LINKED ISSUE #207: Multi gpu strategy enabled even in 1 GPU which possibly not ideal
Multi gpu strategy defined here for CNN https://github.com/wellcometrust/WellcomeML/blob/493e977e360f91d004acaf7cd7fd88bcb0171801/wellcomeml/ml/cnn.py#L199

is enabled even for 1 GPU something that may create an overhead so let's enable it only if there are multiple GPUs.

nsorros:
Description
---

Enable multi gpu strategy when more than 1 GPUs are present.

Fixes #207 

Checklist
---

- [ ] Added tests

### DIFF
--- a/wellcomeml/ml/bilstm.py
+++ b/wellcomeml/ml/bilstm.py
@@ -167,7 +167,7 @@ class BiLSTMClassifier(BaseEstimator, ClassifierMixin):
         steps_per_epoch = math.ceil(X_train.shape[0] / self.batch_size)
         validation_steps = math.ceil(X_val.shape[0] / self.batch_size)
 
-        if tf.config.list_physical_devices('GPU'):
+        if len(tf.config.list_physical_devices('GPU')) > 1:
             strategy = tf.distribute.MirroredStrategy()
         else:
             strategy = tf.distribute.get_strategy()
--- a/wellcomeml/ml/cnn.py
+++ b/wellcomeml/ml/cnn.py
@@ -200,7 +200,7 @@ class CNNClassifier(BaseEstimator, ClassifierMixin):
         steps_per_epoch = math.ceil(X_train.shape[0] / self.batch_size)
         validation_steps = math.ceil(X_val.shape[0] / self.batch_size)
 
-        if tf.config.list_physical_devices('GPU'):
+        if len(tf.config.list_physical_devices('GPU')) > 1:
             strategy = tf.distribute.MirroredStrategy()
         else:  # use default strategy
             strategy = tf.distribute.get_strategy()


wellcomeml/ml/bilstm.py
```diff
@@ -163,7 +163,7 @@ def fit(self, X, Y, embedding_matrix=None, *_):
         steps_per_epoch = math.ceil(X_train.shape[0] / self.batch_size)
         validation_steps = math.ceil(X_val.shape[0] / self.batch_size)
 
-        if tf.config.list_physical_devices('GPU'):
+        if len(tf.config.list_physical_devices('GPU')) > 1:
```
lizgzil, line 166: will this error if `tf.config.list_physical_devices('GPU')` doesnt exist, or are you confident this will never happen?

aCampello:
That's good. Can you add to semantic similarity too please !

wellcomeml/ml/bilstm.py
nsorros: i tried on my CPU and in that case it runs an empty list

nsorros:
> That's good. Can you add to semantic similarity too please !

You mean add multi gpu training?

aCampello:
> You mean add multi gpu training?

Yes,

I think the same line of code can go in the the fit function of Semantic Bert

nsorros:
> > You mean add multi gpu training?
> 
> Yes,
> 
> I think the same line of code can go in the the fit function of Semantic Bert

Sure will give it a go

aCampello (approved):
You'll have a merge conflict, but otherwise looks good to me.

wellcomeml/ml/bert_semantic_equivalence.py
```diff
@@ -181,7 +188,9 @@ def fit(self, X, y, random_state=None, epochs=3, metrics=[], **kwargs):
             )
             self.valid_dataset = self.valid_dataset.batch(self.eval_batch_size)
 
-            self._compile_model()
+            strategy = self._get_distributed_strategy()
```
aCampello, line 191: Sorry for that, but I spoke too soon. Just tested ant this won't work, because you need to create the model with this strategy.

So you need to put `strategy = self._get_distributed_strategy()` before `self.initialise_models()` (L172).

aCampello: And than that line will become

```python
with strategy.scope(): 
     self.initialise_models()
```

wellcomeml/ml/bert_semantic_equivalence.py
```diff
@@ -181,7 +188,9 @@ def fit(self, X, y, random_state=None, epochs=3, metrics=[], **kwargs):
...
+            with strategy.scope():
```
aCampello, line 192: This is correct.

aCampello (changes requested):
Sorry Nick, I just tested and the strategy scope has to be initialised a bit sooner. Happy to push the changes if you prefer.

wellcomeml/ml/bert_semantic_equivalence.py
nsorros: makes sense, i did that when loading but not here. ok will change

aCampello:
The predict function stopped working for me.

```
/data/datalabs/modelling/grants/grant-inference/build/virtualenv/lib/python3.7/site-packages/tensorflow/python/framework/constant_op.py in convert_to_eager_tensor(value, ctx, dtype)
     96       dtype = dtypes.as_dtype(dtype).as_datatype_enum
     97   ctx.ensure_initialized()
---> 98   return ops.EagerTensor(value, ctx.device_name, dtype)
     99 
    100 

ValueError: Attempt to convert a value (None) with an unsupported type (<class 'NoneType'>) to a Tensor
```
Wonder if it works for your other models..

nsorros:
> The predict function stopped working for me.
> 
> ```
> /data/datalabs/modelling/grants/grant-inference/build/virtualenv/lib/python3.7/site-packages/tensorflow/python/framework/constant_op.py in convert_to_eager_tensor(value, ctx, dtype)
>      96       dtype = dtypes.as_dtype(dtype).as_datatype_enum
>      97   ctx.ensure_initialized()
> ---> 98   return ops.EagerTensor(value, ctx.device_name, dtype)
>      99 
>     100 
> 
> ValueError: Attempt to convert a value (None) with an unsupported type (<class 'NoneType'>) to a Tensor
> ```
> 
> Wonder if it works for your other models..

do you want to split the addition to bert classifier to another PR that you can keep working on? this was intended to do a simple change to the CNN so that it does not trigger the multi GPU strategy if there is only 1 GPU.

aCampello:
Yes, there seems to be a problem with predicting for the Multi-GPU, probably because of transformers. Happy to break it down and implement the change appropriately.

nsorros:
Multi GPU for Bert equivalence classifier moved to separate PR #216

aCampello (approved):

Issue #197, opened 2021-02-04
HDBScan numpy C header error


nsorros:
Reproducing the error.
```
pip install numpy==1.19.2 hdbscan
python -c "import hdbscan"
```
throws
```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/Users/nicksorros/code/test/pip_problem/venv/lib/python3.7/site-packages/hdbscan/__init__.py", line 1, in <module>
    from .hdbscan_ import HDBSCAN, hdbscan
  File "/Users/nicksorros/code/test/pip_problem/venv/lib/python3.7/site-packages/hdbscan/hdbscan_.py", line 21, in <module>
    from ._hdbscan_linkage import (single_linkage,
  File "hdbscan/_hdbscan_linkage.pyx", line 1, in init hdbscan._hdbscan_linkage
ValueError: numpy.ndarray size changed, may indicate binary incompatibility. Expected 88 from C header, got 80 from PyObject
```

This is issue is mentioned in hdbscan issue tracker https://github.com/scikit-learn-contrib/hdbscan/issues/457

nsorros:
Our current workaround is to disable hdbscan from the clustering module.

nsorros:
There is a potential solution in the issue which is installing hdbscan separately through `pip install hdbscan --no-cache-dir --no-binary :all: --no-build-isolation`

lizgzil:
@nsorros do you mean similar to this bit https://github.com/wellcometrust/WellcomeML/blob/493e977e360f91d004acaf7cd7fd88bcb0171801/wellcomeml/ml/sent2vec_vectorizer.py#L16

```
        try:
            import sent2vec
        except ImportError:
            from wellcomeml.__main__ import download

            download("non_pypi_packages")
            import sent2vec
```
?
I like that way. If they havent built build/virtualenv then I don't think this won't work anyway. So not sure if it will mess up another venv they might be using?

nsorros:
> @nsorros do you mean similar to this bit
> 
> https://github.com/wellcometrust/WellcomeML/blob/493e977e360f91d004acaf7cd7fd88bcb0171801/wellcomeml/ml/sent2vec_vectorizer.py#L16
> 
> ```
>         try:
>             import sent2vec
>         except ImportError:
>             from wellcomeml.__main__ import download
> 
>             download("non_pypi_packages")
>             import sent2vec
> ```
> 
> ?
[...]

PR #219, opened 2021-02-17
Fix error caught when hdbscan is missing

### LINKED ISSUE #217: Wellcomeml ml imports broken
In the latest master, but not in the latest release, after merging #203, imports of any WellcomeML module are broken because `hdbscan` is missing. There are two problems here:

- We do not throw the useful message to the user that we designed, because this is thrown when a user imports explicitly the clustering module
- The intended change was not suppose to affect other modules, the intended behaviour was for the error to be thrown only if a user tries to import the clustering class

```
>>> from wellcomeml.ml import CNNClassifier
2021-02-16 20:38:00 wellcomeml.logger ERROR: No module named 'hdbscan'
2021-02-16 20:38:00 wellcomeml.logger WARNING: Using WellcomeML without extras (transformers & torch).
ImportError: cannot import name 'CNNClassifier' from 'wellcomeml.ml' (/Users/emiliahadjikyriakou/code/test/WellcomeML/wellcomeml/ml/__init__.py)
```

nsorros:
Description
---

Fixes #217

Essentially we were catching a `ValueError` instead of `ModuleNotFoundError`

Checklist
---

- [ ] Added tests

### DIFF
--- a/wellcomeml/ml/clustering.py
+++ b/wellcomeml/ml/clustering.py
@@ -15,7 +15,7 @@ from sklearn.pipeline import Pipeline
 try:
     from hdbscan import HDBSCAN
     HDBSCAN_INSTALLED = True
-except ValueError:
+except ModuleNotFoundError:
     HDBSCAN_INSTALLED = False
     logger.warning(
         "If you want to use hdbscan you need to run"
--- a/wellcomeml/ml/clustering.py
+++ b/wellcomeml/ml/clustering.py
@@ -15,7 +15,7 @@ from sklearn.pipeline import Pipeline
 try:
     from hdbscan import HDBSCAN
     HDBSCAN_INSTALLED = True
-except ModuleNotFoundError:
+except (ValueError, ModuleNotFoundError):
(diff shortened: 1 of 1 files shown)


wellcomeml/ml/clustering.py
```diff
@@ -15,7 +15,7 @@
 try:
     from hdbscan import HDBSCAN
     HDBSCAN_INSTALLED = True
-except ValueError:
+except ModuleNotFoundError:
```
aCampello, line 18: I think we need to catch both errors actually.

If hdbscan is installed (but not installed with the right flags), it will still throw a value error as per https://github.com/wellcometrust/WellcomeML/issues/197. 

So maybe `except (ValueError, ModuleNotFoundError)`.

aCampello (changes requested):

wellcomeml/ml/clustering.py
nsorros: 394062a

aCampello (approved):

PR #226, opened 2021-02-23
Fix BertClassifier load

### LINKED ISSUE #225: BertClassifier load throws AttributeError
Code to reproduce
```
from wellcomeml.ml import BertClassifier

model = BertClassifier()
model.load("bert") # param does not matter
```

throws
```
  File "/Users/nsorros/code/WellcomeML/wellcomeml/ml/bert_classifier.py", line 191, in load
    self.init_model()
AttributeError: 'BertClassifier' object has no attribute 'init_model'
```

nsorros:
Description
---

Fixes #225

Checklist
---

- [x] Added tests

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -19,7 +19,7 @@ jobs:
       - name: Setup Python
         uses: actions/setup-python@v1
         with:
-          python-version: 3.7.9
+          python-version: 3.7.10
           architecture: x64
       - name: Checkout Master
         uses: actions/checkout@master
--- a/wellcomeml/ml/bert_classifier.py
+++ b/wellcomeml/ml/bert_classifier.py
@@ -188,5 +188,5 @@ class BertClassifier(BaseEstimator, TransformerMixin):
 
     def load(self, model_path):
         """Loads model from directory model_path. Tokenizer initialised from param pretrained"""
-        self.init_model()
+        self._init_model()
         self.model = TFBertForSequenceClassification.from_pretrained(model_path)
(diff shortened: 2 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -19,7 +19,7 @@ jobs:
       - name: Setup Python
         uses: actions/setup-python@v1
         with:
-          python-version: 3.7.9
+          python-version: 3.7.10
```
aCampello, line 22: interesting, just 3.7 or 3.8 won't do?

aCampello (approved):
Nice test. LGTM, assuming it passes/runs on travis 🤞

wellcomeml/ml/bert_classifier.py
```diff
@@ -188,5 +188,5 @@ def save(self, model_path):
 
     def load(self, model_path):
         """Loads model from directory model_path. Tokenizer initialised from param pretrained"""
-        self.init_model()
+        self._init_model()
```
lizgzil, line 191: this is a bit duplicated because in _init_model() you also do `TFBertForSequenceClassification.from_pretrained()`?

lizgzil: if the loading takes a long time then I guess you might just want to take parts of _init_model() rather than running it all?

.github/workflows/main.yml
nsorros: didn't try :p

lizgzil (approved):
looks good and tests pass

PR #228, opened 2021-02-23
Pin dev python version to 3.8

### LINKED ISSUE #224: Explicit python version in Makefile
At the moment we use `python3` as our python version in Makefile
https://github.com/wellcometrust/WellcomeML/blob/9acf7cea9641000980e5bb4bfe80daebc5c5e5c9/Makefile#L4

This is not reproducible as each of us including travis might be running on a different python. We force travis to try python 3.7 and 3.8 which is a good thing but we should also ensure our env is created with one version.

An added benefit of this is that brew cannot mess up your python as python3.8 should link to one version compared to python3 that might be updated to point to the latest.

nsorros:
Description
---

Fixes #224

Checklist
---

- [ ] Added tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 .DEFAULT_GOAL := all
 
 VIRTUALENV := build/virtualenv
-PYTHON_VERSION := python3
+PYTHON_VERSION := python3.8
 
 $(VIRTUALENV)/.installed:
 	@if [ -d $(VIRTUALENV) ]; then rm -rf $(VIRTUALENV); fi


aCampello:
It is going to be tricky to make both environment pass for both pythons. One workaround is to do

```make PYTHON_VERSION=python3 virtualenv``` 

in the `travis.yml` CI, and let the travis environment decide which python3 to use.

Alternatively, you can modify the `install` script to make it depend on the python environment, and then use a similar command as above, but with the specific python.

**Comment by nsorros** [comment]
> It is going to be tricky to make both environment pass for both pythons. One workaround is to do
> 
> `make PYTHON_VERSION=python3 virtualenv`
> 
> in the `travis.yml` CI, and let the travis environment decide which python3 to use.
> 
> Alternatively, you can modify the `install` script to make it depend on the python environment, and then use a similar command as above, but with the specific python.

Weird situation, will do some research as well. Not sure what the best practice is here.

**Comment by jdu** [comment]
You could add an environment variable to your .travis.yaml and assign the python version in the Makefile based on if that flag exists.

for instance you could add 

```
env:
  global:
    - IS_IN_TRAVIS=1
```

Then modify the Makefile to have 

```
ifdef IS_IN_TRAVIS
PYTHON_VERSION := python3
else
PYTHON_VERSION := python3.8
endif
```

nsorros:
I am now having second thoughts on whether this change is necessary. Initially I thought that this moves us towards reproducibility which it does but we are testing and supporting 3.7 and 3.8 at this point so for as long as someone develops with either that should be fine. What is the benefit of restricting development to 3.8, if anything you can argue that this way error with 3.7 will be rarely caught from us.

Thoughts @jdu @aCampello

jdu:
> I am now having second thoughts on whether this change is necessary. Initially I thought that this moves us towards reproducibility which it does but we are testing and supporting 3.7 and 3.8 at this point so for as long as someone develops with either that should be fine. What is the benefit of restricting development to 3.8, if anything you can argue that this way error with 3.7 will be rarely caught from us.
> 
> Thoughts @jdu @aCampello

I mean as long as you avoid assignment expressions and positional-only arguments anything you build for 3.7 should work fine for 3.8 generally. Unless there's an issue with one of the libraries that you're depending on that needs 3.8?

nsorros:
> > I am now having second thoughts on whether this change is necessary. Initially I thought that this moves us towards reproducibility which it does but we are testing and supporting 3.7 and 3.8 at this point so for as long as someone develops with either that should be fine. What is the benefit of restricting development to 3.8, if anything you can argue that this way error with 3.7 will be rarely caught from us.
> > Thoughts @jdu @aCampello
> 
> I mean as long as you avoid assignment expressions and positional-only arguments anything you build for 3.7 should work fine for 3.8 generally. Unless there's an issue with one of the libraries that you're depending on that needs 3.8?

I think the problem was the libraries, probably tensorflow or spacy in particular but not sure. One of them required python 3.7

aCampello:
Tensorflow required 3.7, that's why we allowed failures fro 3.8, but now they improved their support to 3.8 and both pass

nsorros:
> Tensorflow required 3.7, that's why we allowed failures fro 3.8, but now they improved their support to 3.8 and both pass

@jdu suggests that `tox` is the magic bullet we want

nsorros:
> > Tensorflow required 3.7, that's why we allowed failures fro 3.8, but now they improved their support to 3.8 and both pass
> 
> @jdu suggests that `tox` is the magic bullet we want

is there consensus for switching to tox? @aCampello @lizgzil. tox essentially creates a virtualenv and runs the tests, this makes tests more reproducible as you also manage the environment that tests run which now can be different in travis and our machines.

aCampello:
Tox is great, I think Matt used it before for another project. I am happy to go with it!

nsorros:
> Tox is great, I think @ivyleavedtoadflax used it before for another project with us. I am happy to go with it!

Ok will transition us to tox. It is written from the same guy that wrote pytest so it should play nicely.

nsorros:
Closing this and will reopen with tox

Issue #160, opened 2020-10-19
Fix license


aCampello:
At the moment our license is not correctly displayed/picked up by setup.py

https://github.com/wellcometrust/WellcomeML/blob/15df012b4e16d726dfb622099e2f1725fcf8f6ec/setup.py#L37

PR #231, opened 2021-02-24
Release 1.0.1

aCampello:
Release checklist

- [x] Change `wellcomeml/__version__.py`
- [x] Add changelog
- [x] `make dist`
- [x] Verify new package was generated correctly on the [pip registry](https://pypi.org/project/wellcomeml)
 and GitHub releases
- [x] "Yank" CalVER versions on pyPi registry
- [x] Make sure `pip install wellcomeml` install correct version

Fixes https://github.com/wellcometrust/WellcomeML/issues/160


wellcomeml/__version__.py
```diff
@@ -1,5 +1,5 @@
 __name__ = "wellcomeml"
-__version__ = "2021.2.1"
+__version__ = "1.0.0"
```
lizgzil, line 2: 🎉

nsorros (approved):
LGTM, merge when you tick all the boxes

aCampello:
`make dist` works uneventfully

aCampello:
Pypi complained about 1.0.0 

```bash
HTTPError: 400 Bad Request from https://upload.pypi.org/legacy/
This filename has already been used, use a different version. See https://pypi.org/help/#file-name-reuse for more information.
```
Probably something to do with the fact that I used 1.0.0 before for test... investigating

aCampello:
Well this is awkward, seems that PyPI is not enjoying 1.0.0 because there was a test one created in the past. Will figure out if there is anyway to by pass it unless we'll have to be 1.0.1 or something!

aCampello:
Ok, for a number of reasons it seems that we won't be able to release 1.0.0 on pypi, since in the past there was a 1.0.0 deleted.

https://github.com/pypa/packaging-problems/issues/74

nsorros:
> Ok, for a number of reasons it seems that we won't be able to release 1.0.0 on pypi, since in the past there was a 1.0.0 deleted.
> 
> [pypa/packaging-problems#74](https://github.com/pypa/packaging-problems/issues/74)

Yep, releases are not mutable. If they were that would be a problem since one day a library you installed by pinning a version worked and the other it did not out of no fault of your own. 1.0.1 it is.

In the future btw if you want to test something, PyPi recommends doing it in a test account. I also have a test account which I used when we first released wellcomeml

aCampello:
Yes, that was mea culpa. I was too sloppy with the test.

aCampello:
Running make dist again....

aCampello:
Pip install outputs:
```
> pip freeze | grep wellcomeml
wellcomeml==1.0.1
```

aCampello:
Trying to pip install old versions now. Pip installing pinned versions work:

```
pip install wellcomeml==2020.9.0 --no-cache-dir
```

Returns

```
pip freeze | grep wellcomeml
wellcomeml==2020.9.0
```

Note that assignments like `wellcomeml < 2020.9.0` won't work. User will have to pin the exact version.

aCampello:
Incidentally, this PR fixes https://github.com/wellcometrust/WellcomeML/issues/160

PR #233, opened 2021-02-24
Fix CNN predict when X tf.data.Dataset

### LINKED ISSUE #232: CNN predict behaves wrong for input tf.dataDataset
# To reproduce
```
from wellcomeml.ml import CNNClassifier, KerasVectorizer
import numpy as np
import tensorflow as tf

X = [
    "One",
    "One only",
    "Two nothing else",
    "Two and three"
]
Y = np.array([0, 0, 1, 1])

vec = KerasVectorizer()
X_vec = vec.fit_transform(X)

data = tf.data.Dataset.from_tensor_slices((X_vec, Y))
data = data.shuffle(100)
clf = CNNClassifier(batch_size=2)

clf.fit(data)
Y_pred = clf.predict(data)
print(Y_pred.shape)
```

returns

```
tensorflow WARNING: Model was constructed with shape (None, 3) for input KerasTensor(type_spec=TensorSpec(shape=(None, 3), dtype=tf.float32, name='input_1'), name='input_1', description="created by layer 'input_1'"), but it was called on an input with incompatible shape (3, 1).
(12, 1)
```

So it outputs 12 predictions instead of 4

nsorros:
Description
---

Fixes #232

Checklist
---

- [x] Added tests

### DIFF
--- a/tests/test_cnn.py
+++ b/tests/test_cnn.py
@@ -246,4 +246,4 @@ def test_XY_dataset():
     clf = CNNClassifier(batch_size=2)
 
     clf.fit(data)
-    assert clf.score(X_vec, Y) > 0.6
+    assert clf.score(data, Y) > 0.6
--- a/tests/test_cnn.py
+++ b/tests/test_cnn.py
@@ -247,3 +247,30 @@ def test_XY_dataset():
 
     clf.fit(data)
     assert clf.score(data, Y) > 0.6
+
+
+def test_XY_dataset_sparse_y():
+    X = [
+        "One and two",
+        "One only",
+        "Two nothing else",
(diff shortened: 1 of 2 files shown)


tests/test_cnn.py
```diff
@@ -242,8 +242,35 @@ def test_XY_dataset():
     X_vec = vec.fit_transform(X)
 
     data = tf.data.Dataset.from_tensor_slices((X_vec, Y))
-    data = data.shuffle(100)
+    data = data.shuffle(100, seed=42)
```
aCampello, line 245: The famous seed

aCampello (approved):
LGTM if pipelines pass

PR #235, opened 2021-02-25
Release 1.0.2

aCampello:
- [x] Change `wellcomeml/__version__.py`
- [x] Add changelog
- [x] `make dist`
- [x] Verify new package was generated correctly on the [pip registry](https://pypi.org/project/wellcomeml)
 and GitHub releases

### DIFF
--- a/create_release.sh
+++ b/create_release.sh
@@ -1,4 +1,6 @@
 #!/bin/sh
+git fetch && git rebase origin/master
+
 VERSION=$(python setup.py --version)
 VIRTUALENV=build/virtualenv
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,16 @@
+# 1.0.2
+Release date: 25/02/2021
+
+Due to an error, some of the improvements in the previous release weren't built in the PyPI whl.
+
+- #233 CNN predict when X tf.data.Dataset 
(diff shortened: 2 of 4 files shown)


nsorros (approved):
LGTM 👍

create_release.sh
```diff
@@ -1,4 +1,6 @@
 #!/bin/sh
+git fetch && git rebase origin/master
```
nsorros, line 2: maybe you want to ensure you are in master. an alternative is to change that to `git fetch && git reset --hard origin/master` in which case it won't matter.

aCampello: I don't it! I'm pushing for `release-1.0.2`, I think if I reset-hard I'll lose all the commits from this branch

aCampello: Ah, I see what you mean. I could alternatively checkout to master? 

```git fetch && git checkout/master && git reset --hard origin/master```?

nsorros: I thought you were doing it from master, which makes no sense. Leave as is.

aCampello:
```pip install wellcomeml --no-cache-dir``` outputs the new version. Just for a sanity check, the HDbscan bug is gone, tested using

``` python -c "from wellcomeml.ml import CNNClassifier"```

PR #237, opened 2021-02-25
Fix keras vectorizer by inferring sequence length from data

### LINKED ISSUE #236: KerasVectorizer transform returns sequences of various lengths
# To reproduce

```
from wellcomeml.ml import KerasVectorizer

texts = ["This is a sentence", "Another"]
vectorizer = KerasVectorizer.fit(texts)
X_vec = vectorizer.transform(texts)
print(X_vec.shape)
# returns 

texts.= ["This"]
X_vec = vectorizer.transform(texts)
print(X_vec.shape)
# returns

texts.= ["This is a slightly longer sentence, what happens?"]
X_vec = vectorizer.transform(texts)
print(X_vec.shape)
# returns 
```

This means that train and test sequence length is not the same. This is something that older version of tensorflow keras tolerated (see below) possibly by automatically padding or truncating but the latest version throws an error.

This can be fixed on the user end if they specific the sequence length as an argument in the KerasVectorizer as in the case KerasVectorizer respects that length so the problem is only in the case that the user does not pass the argument. In the case I suspect that the intended behaviour expected is for the vectorizer to set the variable from the data.

# Tensorflow versions investigation

```
from wellcomeml.ml import KerasVectorizer, CNNClassifier
import numpy as np

texts = [
    "This is a train sentence",
    "which means it will be used for training",
    "but not for testing"
]

vectorizer = KerasVectorizer()
vectorizer.fit(texts)

X_vec = vectorizer.transform(texts)

model = CNNClassifier()
model.fit(X_vec, np.array([0,0,1]))

test_texts = ["This is a test sentence that will be longer than usual to test if there is a bug"]
X_vec_test = vectorizer.transform(test_texts)
model.predict(X_vec_test)
```

in `wellcomeml==2011.11.1` which installs `tensorflow==2.3.2` runs ok
in `wellcomeml==1.0.1` which install `tensorflow==2.4.0` it throws `ValueError: Input 0 is incompatible with layer model: expected shape=(None, 8), found shape=(None, 18)`

nsorros:
Description
---

Fixes #236

Checklist
---

- [x] Added tests

### DIFF
--- a/wellcomeml/ml/keras_vectorizer.py
+++ b/wellcomeml/ml/keras_vectorizer.py
@@ -44,8 +44,8 @@ class KerasVectorizer(BaseEstimator, TransformerMixin):
         max_sequence_length = 1
 
         def update_max_sequence_length(X_buffer, max_sequence_length, load_buffer):
-            X_vec = self.transform(X_buffer)
-            sequence_length = X_vec.shape[1]
+            X_tokens = self.tokenizer.encode(X_buffer)
+            sequence_length = max([len(x) for x in X_tokens])
             if sequence_length >= max_sequence_length:
                 max_sequence_length = sequence_length
             return max_sequence_length
--- a/wellcomeml/ml/keras_vectorizer.py
+++ b/wellcomeml/ml/keras_vectorizer.py
@@ -10,6 +10,7 @@ import gensim.downloader as api
 
 from os import path
 
+from wellcomeml.logger import logger
 from wellcomeml.ml.transformers_tokenizer import TransformersTokenizer
 
 
@@ -38,6 +39,30 @@ class KerasVectorizer(BaseEstimator, TransformerMixin):
         self.sequence_length = sequence_length
         self.tokenizer_library = tokenizer_library
 
+    def _infer_from_data(self, X, load_buffer=1000):
+        # We could look at a sample for more efficient
+        max_sequence_length = 1
+
+        X_buffer = []
+        for x in X:
+            X_buffer.append(x)
+
+            if len(X_buffer) >= load_buffer:
+                X_vec = self.transform(X_buffer)
+                sequence_length = X_vec.shape[1]
+                if sequence_length >= max_sequence_length:
+                    max_sequence_length = sequence_length
+                X_buffer = []
+
+        if X_buffer:
+            # TODO: Refactor
+            X_vec = self.transform(X_buffer)
+            sequence_length = X_vec.shape[1]
+            if sequence_length >= max_sequence_length:
+                max_sequence_length = sequence_length
+
+        self.sequence_length = max_sequence_length
+
     def fit(self, X, *_):
         if self.tokenizer_library == "keras":
             self.tokenizer = KerasTokenizer(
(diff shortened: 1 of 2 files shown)


wellcomeml/ml/keras_vectorizer.py
```diff
@@ -38,6 +39,32 @@ def __init__(self, vocab_size=None, sequence_length=None, oov_token="<OOV>",
         self.sequence_length = sequence_length
         self.tokenizer_library = tokenizer_library
 
+    def _infer_from_data(self, X, load_buffer=1000):
+        # We could look at a sample for more efficient
+        max_sequence_length = 1
+
+        def update_max_sequence_length(X_buffer, max_sequence_length, load_buffer):
+            X_vec = self.transform(X_buffer)
```
lizgzil, line 47: If time is a big problem with this, then I wonder if there is something you can do to not have to transform every `X_buffer` (I assume this is the costly bit). 
For example, if `sum([len(x) for x in X_buffer]) < max_sequence_length` (number of words in all sentences not removing duplicates) then maybe it can be assumed that `X_vec.shape[1] > max_sequence_length` can't occur?

Perhaps `sum([len(x) for x in X_buffer]) < max_sequence_length` never holds though.

lizgzil (approved):
looks fine to me, just a possible time saving suggestion

wellcomeml/ml/keras_vectorizer.py
nsorros: len(x) is the number of characters though. the expensive bit is tokenisation, going from text to tokens which you cannot avoid. the only thing that can go faster is padding which is not strictly necessary but it performs exactly the same operation I implement under the hood, maybe faster maybe slower.

note that transform is essentially a two lines function
```
sequences = self.tokenizer.encode(X)
return pad_sequences(sequences, maxlen=self.sequence_length)
```

I will investigate a bit the timing before I merge to see whether there is a cost saving I could do 😬

nsorros: So created this script

```
import json
import time

from tensorflow.keras.preprocessing.sequence import pad_sequences
import tensorflow as tf


DATA_PATH = "sample_disease_mesh.jsonl"


def yield_texts(DATA_PATH):
    with open(DATA_PATH) as f:
        for line in f:
            item = json.loads(line)
            yield item["text"]

tokenizer = tf.keras.preprocessing.text.Tokenizer()
tokenizer.fit_on_texts(yield_texts(DATA_PATH))

texts = list(yield_texts(DATA_PATH))
```
Version 1
```
start = time.time()
seqs = tokenizer.texts_to_sequences(texts)
pad_seqs = pad_sequences(seqs)
print(pad_seqs.shape[1])
print(time.time()-start)
```

Version 2
```
start = time.time()
seqs = tokenizer.texts_to_sequences(texts)
pad_seqs = pad_sequences(seqs)
print(pad_seqs.shape[1])
print(time.time()-start)
```

Results
* Version 1: 2.3s
* Version 2: 2s

So there is a slight speedup so will add the faster version in 🏎️

lizgzil: glad youve found a speed up! v1 and v2 looks the same in this

PR #244, opened 2021-03-11
Loosen spacy requirements to less than 3.1.0 and more than 3.0.0

### LINKED ISSUE #243: Spacy dependency conflict with SciSpacy and is too strict
Our current requirement for spacy is `spacy[lookups]==3.0.0` which is too strict and it creates a conflict with scispacy which asks for >3.0.1 due to a possible fix that is introduced. We can definitely loosen the dependency slighly without breaking anything hopefully.

nsorros:
Description
---

Fixes #243

Checklist
---

- [ ] Added tests

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -46,7 +46,7 @@ setuptools.setup(
         'boto3',
         'scikit-learn',
         'scipy==1.4.1',
-        'spacy[lookups]<3.1.0,>=3.0.0',
+        'spacy[lookups]<4.0.0,>=3.0.0',
         'umap-learn',
         'nervaluate',
         'twine',
--- a/setup.py
+++ b/setup.py
@@ -46,7 +46,7 @@ setuptools.setup(
         'boto3',
         'scikit-learn',
         'scipy==1.4.1',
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -46,7 +46,7 @@
         'boto3',
         'scikit-learn',
         'scipy==1.4.1',
-        'spacy[lookups]==3.0.0',
+        'spacy[lookups]<3.1.0,>=3.0.0',
```
aCampello, line 49: In the issue description, it says that scispacy which asks for `>3.0.1.`. Do you think just `> 3.0.1` wouldn't work?

aCampello (approved):
It all passes, go ahead and merge! But I was wondering whether we can relax even more ?

setup.py
nsorros: It would but I guess we should set an upper bound, should we say <4.0.0?

aCampello: Yeah, that'd be more relaxed!

nsorros: aa2dc2c

PR #245, opened 2021-03-11
VotingClassifier extras

lizgzil:
Description
---
https://trello.com/c/QiOxLF9l/1195-additions-to-wellcomevotingclassifier

- Add ability to input list of tuples to voting classifier - the estimator and a vectoriser
- In the binary + hard case, you can specify a num_agree parameter to custom how many number of models need to agree in order for final score to be included. before: >50% models agree, custom: >=x models agree. if left unspecified this will default to majority rule (same as before), where if there are 3 models, >=2 need to agree, and if there are 4 models >=3 need to agree.

- New tests for different scenarios

- docstring explaining the nuances

Checklist
---

- [x] Added tests

### DIFF
--- a/tests/test_voting_classifier.py
+++ b/tests/test_voting_classifier.py
@@ -31,9 +31,10 @@ class MockEstimator():
 class MockVectorizer():
     def __init__(self, size):
         self.size = size
-        
+
     def transform(self, X):
         return np.random.rand(len(X), self.size)
+
 
 def test_multilabel():
     Y1_prob = np.array([
@@ -258,4 +259,3 @@ def test_binary_hard_num_agree():
     X = ["mock", "data", "not used"]
     Y = voting_classifier.predict(X)
     assert np.array_equal(Y, Y_expected)
-
--- a/tests/test_voting_classifier.py
+++ b/tests/test_voting_classifier.py
@@ -153,37 +153,76 @@ def test_hard_voting():
     Y = voting_classifier.predict(X)
     assert np.array_equal(Y, Y_expected)
 
-
-def test_inc_vectorizer():
-
+def test_multilabel_hard():
+    # Since no num_agree, majority (>=2/3) agreement needed
     Y1_prob = np.array([
-        [0.9, 0.1],
-        [0.2, 0.8],
-        [0.3, 0.7]
+        [0.9, 0.9],
+        [0.1, 0.9],
+        [0.1, 0.9]
     ])
     Y2_prob = np.array([
-        [0.7, 0.3],
+        [0.9, 0.9],
+        [0.9, 0.9],
+        [0.1, 0.1]
+    ])
+    Y3_prob = np.array([
+        [0.9, 0.9],
         [0.9, 0.1],
-        [0.4, 0.6]
+        [0.1, 0.1]
     ])
     Y_expected = np.array([
-        0,
-        0,
-        1
+        [1, 1],
+        [1, 1],
+        [0, 0]
     ])
-
-    pipe1 = Pipeline([('vect', MockVectorizer(size=3)), ('est', MockEstimator(Y1_prob))])
-    pipe2 = Pipeline([('vect', MockVectorizer(size=5)), ('est', MockEstimator(Y2_prob))])
+    est1 = MockEstimator(Y1_prob, multilabel=True)
+    est2 = MockEstimator(Y2_prob, multilabel=True)
+    est3 = MockEstimator(Y3_prob, multilabel=True)
 
     voting_classifier = WellcomeVotingClassifier(
-        estimators=[pipe1, pipe2], voting="soft"
+        estimators=[est1, est2, est3], voting="hard", multilabel=True
     )
     X = ["mock", "data", "not used"]
     Y = voting_classifier.predict(X)
     assert np.array_equal(Y, Y_expected)
 
+def test_multilabel_hard_num_agree():
+    # >=3/3 agreement needed
+    Y1_prob = np.array([
+        [0.9, 0.9],
+        [0.1, 0.9],
+        [0.1, 0.9]
+    ])
+    Y2_prob = np.array([
+        [0.9, 0.9],
+        [0.9, 0.9],
+        [0.1, 0.1]
+    ])
+    Y3_prob = np.array([
+        [0.9, 0.9],
+        [0.9, 0.1],
+        [0.1, 0.1]
+    ])
+    Y_expected = np.array([
+        [1, 1],
+        [0, 0],
+        [0, 0]
+    ])
+    est1 = MockEstimator(Y1_prob, multilabel=True)
+    est2 = MockEstimator(Y2_prob, multilabel=True)
+    est3 = MockEstimator(Y3_prob, multilabel=True)
 
-def test_binary_hard_majority():
+    voting_classifier = WellcomeVotingClassifier(
+        estimators=[est1, est2, est3],
+        voting="hard",
+        multilabel=True,
+        num_agree=3
+    )
+    X = ["mock", "data", "not used"]
+    Y = voting_classifier.predict(X)
+    assert np.array_equal(Y, Y_expected)
+
+def test_binary_hard():
 
     Y1_prob = np.array([
         [0.9, 0.1],
@@ -247,7 +286,7 @@ def test_binary_hard_num_agree():
     ])
     Y_expected = np.array([
         0,
-        1,
+        0,
         1
     ])
     est1 = MockEstimator(Y1_prob)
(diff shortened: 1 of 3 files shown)


aCampello:
Now we are talking! :D Some flake8 issues, but super exciting.

wellcomeml/ml/voting_classifier.py
```diff
@@ -29,30 +30,59 @@ def _get_estimators(self):
...
         if self.pretrained:
-            check_is_fitted(self, 'estimators')
+            check_is_fitted(self, "estimators")
 
             estimators = self._get_estimators()
+            has_vectorizers = self._check_vectorizer(estimators)
 
             if self.voting == "soft":
-                Y_probs = np.array([est.predict_proba(X) for est in estimators])
+                if has_vectorizers:
```
nsorros, line 47: why do you add that? I am under the impression it is not necessary. an estimator can be a classifier or a pipeline that contains a vectorizer and the code should work i.e. when you call predict_proba it runs transform and then predict

wellcomeml/ml/voting_classifier.py
```diff
@@ -29,30 +30,59 @@ def _get_estimators(self):
...
-                    Y_preds = np.column_stack(Y_preds)
-                    axis = 1
-                return np.apply_along_axis(
-                    lambda x: np.argmax(np.bincount(x)),
-                    axis=axis,
-                    arr=Y_preds)
+                    # If num_agree isn't set then use majority vote
+                    # TO DO: a message when if num_agree is specified, but
+                    # voting=soft and/or multilabel=True then num_agree isn't used
+                    # 'num_agree specified but not used - only used in binary hard voting cases'
```
nsorros, line 78: I suggest you catch that and add a warning using logger.

wellcomeml/ml/voting_classifier.py
```diff
@@ -29,30 +30,59 @@ def _get_estimators(self):
         else:  # tuple with named estimators
             return [est for _, est in self.estimators]
 
+    def _check_vectorizer(self, estimators):
+        if type(estimators[0]) == tuple:
```
nsorros, line 34: this does not mean there is a vectorizer

nsorros (changes requested):
I think the vectorizer check is not needed, estimators should work fine with pipelines that contain vectorisers or classifier.

nsorros:
Super excited about this contribution @lizgzil 🎉 Let's iron some details of the implementation before merging.

lizgzil:
@nsorros yes good point, I hadnt been thinking about the input as a pipeline! will update :)

lizgzil:
@nsorros This has now been updated to use pipeline inputs rather than tuples for the case of including a vectorizer, see what you think

wellcomeml/ml/voting_classifier.py
```diff
@@ -31,28 +34,38 @@ def _get_estimators(self):
...
                     return np.argmax(Y_prob, axis=1)
             else:  # hard voting
+
                 Y_preds = [est.predict(X) for est in estimators]
+                Y_preds = np.array(Y_preds)
                 if self.multilabel:
-                    Y_preds = np.array(Y_preds)
-                    axis = 0
+                    if self.num_agree:
+                        logger.warning("num_agree specified but not needed for multilabel voting")
```
nsorros, line 56: is it not needed or not used?

wellcomeml/ml/voting_classifier.py
```diff
@@ -31,28 +34,38 @@ def _get_estimators(self):
...
-                    lambda x: np.argmax(np.bincount(x)),
-                    axis=axis,
-                    arr=Y_preds)
+                    # If num_agree isn't set then use majority vote
+                    if not self.num_agree:
+                        # So if 4 estimators, >= 3 need to agree
+                        self.num_agree = np.ceil((len(estimators) + 1) / 2)
+
+                    return np.apply_along_axis(
+                        lambda x: int(np.sum(x) >= self.num_agree), axis=0, arr=Y_preds
```
nsorros, line 67: I assume this work for binary, does it also work for multi class?

nsorros: For `multilabel` I think something simple like np.array(Y.sum(axis=0) > self.num_agree, dtype=int32) should work.

For `multi class` it is slightly more complicated by here is a two step solution
```
votes = np.apply_along_axis(lambda x: max(np.bincount(x)), axis=0, arr=Y_preds)
max_class = np.apply_along_axis(lambda x: np.argmax(np.bincount(x)), axis=0, arr=Y_preds)
return max_class * (votes > self.num_agree)
```

nsorros: I think that the multi class version above also covers binary

nsorros (approved):
It would be nice to generalise the new feature to multi class and multilabel

lizgzil:
@nsorros for multiclass I wasn't sure if it would work.
e.g. if you have 3 models, one predicts class 0, one predicts class 1, and one predicts class 2. What should the output be? Nan?

lizgzil:
[Additions to WellcomeVotingClassifier](https://trello.com/c/QiOxLF9l/1195-additions-to-wellcomevotingclassifier)

nsorros:
> @nsorros for multiclass I wasn't sure if it would work.
> e.g. if you have 3 models, one predicts class 0, one predicts class 1, and one predicts class 2. What should the output be? Nan?

what is the output of VotingClassifier in that case? I think it is random or always selecting the first of the classes that have the same votes. I think the same happens here, they will all get the same votes and argmax will choose the first.

in the case where self.num_agree is set and no class passes the threshold, the same problem occurs and again the first class will be returned out of all classes if no class passes the threshold or in a tie.

if in doubt we can write some sample code to test this, want me to do that? I think the first thing is to agree what is the behaviour we want and then to ensure we have it implemented. we can also ensure it works as expected with a test.

lizgzil:
@nsorros it seems like VotingClassifier indeed outputs the first class (in numerical order) with the same votes. i.e. if your classes with the same votes are 3,2,4 then 2 will be outputted

nsorros:
> @nsorros it seems like VotingClassifier indeed outputs the first class (in numerical order) with the same votes. i.e. if your classes with the same votes are 3,2,4 then 2 will be outputted

let's maybe try to keep this behaviour and document it.

wellcomeml/ml/voting_classifier.py
lizgzil: these were v useful :) 
In the multiclass option I changed it so that if there is no maximum over the threshold, then it wont default to 0 (as `max_class * (votes > self.num_agree)` does) but rather pick the first value from the classes which were voted on.
e.g. if classes  5, 2, and 3 were voted on twice by different estimators, but `num_agree` was 3, then the output would be class 2.

tests/test_voting_classifier.py
```diff
@@ -261,3 +300,108 @@ def test_binary_hard_num_agree():
...
+    voting_classifier = WellcomeVotingClassifier(
+        estimators=[est1, est2, est3],
+        voting="hard",
+        num_agree=3
+    )
+    X = ["mock", "data", "not used"]
+    Y = voting_classifier.predict(X)
+    assert np.array_equal(Y, Y_expected)
+
+def test_inc_vectorizer():
```
nsorros, line 380: took me some seconds to realise what was the test here. maybe rename to `test_pipeline_with_vectorizer`

wellcomeml/ml/voting_classifier.py
```diff
@@ -49,23 +77,22 @@ def predict(self, X):
...
-                        self.num_agree = np.ceil((len(estimators) + 1) / 2)
-
-                    return np.apply_along_axis(
-                        lambda x: int(np.sum(x) >= self.num_agree), axis=0, arr=Y_preds
-                    )
+                    votes = np.apply_along_axis(lambda x: max(np.bincount(x)), axis=0, arr=Y_preds)
+                    max_class = np.apply_along_axis(lambda x: np.argmax(np.bincount(x)), axis=0, arr=Y_preds)
+                    # If no maximum over the threshold, then pick the first from an ordered list
+                    # of the other options, e.g. if 5,2,3 were voted on pick 2 (not 0)
+                    options = np.sort(np.transpose(Y_preds))
```
nsorros, line 94: options might be a bit confusing as a name. also it took me some time to get my head around what Y_preds contains at this point and why you need to transpose, maybe a small comment to say about the dimensions of Y_preds at this point e.g. Y_preds: n_estimators x n_examples with predicted class as values

nsorros:
It would be nice to start using this at your tech project as well @lizgzil this way you will make sure this is updated and works as expected

PR #246, opened 2021-03-18
Use tox for tests

### LINKED ISSUE #224: Explicit python version in Makefile

nsorros:
Description
---

Fixes #224

Checklist
---

- [x] Added tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 .DEFAULT_GOAL := all
 
 VIRTUALENV := build/virtualenv
-PYTHON_VERSION := python3
+PYTHON_VERSION := python3.8
 
 $(VIRTUALENV)/.installed:
 	@if [ -d $(VIRTUALENV) ]; then rm -rf $(VIRTUALENV); fi
--- a/tox.ini
+++ b/tox.ini
@@ -7,5 +7,5 @@ deps =
 	.[deep-learning]
 
 commands = python -m spacy download en_core_web_sm
-		   pytest -m  "not (integration)" -s -v --durations=0 --disable-warnings --tb=line --cov=wellcomeml ./tests
+           pytest -m  "not (integration)" -s -v --durations=0 --disable-warnings --tb=line --cov=wellcomeml ./tests
 
--- a/tox.ini
+++ b/tox.ini
@@ -1,0 +1,11 @@
+[tox]
+envlist = py37, py38
+
+[testenv]
+deps = 
+	-r requirements_test.txt
+	.[deep-learning]
+
+commands = python -m spacy download en_core_web_sm
+		   pytest -m  "not (integration)" -s -v --durations=0 --disable-warnings --tb=line --cov=wellcomeml ./tests
+
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,5 @@
+dist: bionic
+
 language: python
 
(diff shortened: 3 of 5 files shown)


nsorros:
[Switch WellcomeML to tox](https://trello.com/c/Muz3FagU/1191-switch-wellcomeml-to-tox)

nsorros:
As part of this PR we could potentially switch to Github actions, opened an issue for discussion #247. @lizgzil @aCampello

nsorros:
New tool, new problems. Python3.7 with tox complains about binary incompatibility when installing spacy models. To make things more interestingly, it runs fine locally, a problem that tox should not have as it runs tests in its own isolated environment which is more reproducible than what we were doing before.

Makefile
```diff
@@ -1,7 +1,7 @@
 .DEFAULT_GOAL := all
 
 VIRTUALENV := build/virtualenv
-PYTHON_VERSION := python3
+PYTHON_VERSION := python3.8
```
aCampello, line 4: Any reason why our default is now 3.8? (I'm fine with that btw)

tox.ini
```diff
@@ -0,0 +1,11 @@
+[tox]
+envlist = py37, py38
+
+[testenv]
+deps = 
+	-r requirements_test.txt
+	.[deep-learning]
+
+commands = python -m spacy download en_core_web_sm
+		   pytest -m  "not (integration)" -s -v --durations=0 --disable-warnings --tb=line --cov=wellcomeml ./tests
```
aCampello, line 10: is there an indentation problem here?

aCampello (approved):
Thanks for this change! Looks good!

Makefile
nsorros: Check #224 and #228

tox.ini
nsorros: ini files do not care about indentation so I guess that is why this works but let me see whether this is a space vs tabs issue

Makefile
aCampello: Lol - it was my suggestion to keep 3.8.... 🙈

tox.ini
nsorros: 7888bd292e74f569703ec428a5485c87b1ca4860

nsorros:
On the way we seem to have missed the codecov report/ @aCampello @lizgzil. Investigating.

nsorros:
Open a separate issue for the codecov problem #251

nsorros:
Btw @lizgzil and @aCampello note, that to run tests moving on you need to run `tox` and you need need `python 3.7` and `python 3.8`. If one is missing tests will not run for that version. More importantly, the default python version for development has been switched to `python 3.8` as discussed in #224. If you have any problem installing that version let me know. I have been using `pyenv` to manage multiple python versions which has been working fine.

PR #262, opened 2021-03-25
Add debug stack trace in wellcome/ml.__init__

### LINKED ISSUE #261: `wellcomeml.ml` throws "cryptic" errors in ImportError
Currently if there is an import error in any of our ml packages we catch it and throw a generic message.
https://github.com/wellcometrust/WellcomeML/blob/5f31fac60a6883d1ca9106f68bc48bc4e40bf227/wellcomeml/ml/__init__.py#L33

This was slimly improved by writing the actual error which at least hints as to which import failed but it still provides little information to the user on the source of the problem. And there are a couple of things that might be causing this as we have experienced over time so a bit more help would go a long way.

This is partly related to #257 but it goes beyond as this is not the first time we have been trying to understand the route of that not working. Other examples are #217 #197 #195 #104.

One possible way to improve this is to print the stack trace. We could do that only in debug mode since the way we have implemented this we allow for the use of init without extras. When and if we make deep-learning the default #234 we should just print the stack trace at that point or remove the catching of the exception.

I think in the meantime adding the debug option to print the stack trace will allow us to get more information from what is causing the error. Thoughts @aCampello @lizgzil?

nsorros:
Description
---

Fixes #261

Now when you set `export LOGGING_LEVEL=DEBUG` you can get something like
```
2021-03-25 12:00:07 tensorflow DEBUG: Falling back to TensorFlow client; we recommended you install the Cloud TPU client directly with pip install cloud-tpu-client.
2021-03-25 12:00:12 wellcomeml.logger ERROR: No module named 'torch'
2021-03-25 12:00:12 wellcomeml.logger WARNING: Using WellcomeML without extras (transformers & torch).
2021-03-25 12:00:12 wellcomeml.logger DEBUG: Traceback (most recent call last):
  File "/Users/nsorros/code/WellcomeML/wellcomeml/ml/__init__.py", line 12, in <module>
    from .vectorizer import Vectorizer
  File "/Users/nsorros/code/WellcomeML/wellcomeml/ml/vectorizer.py", line 11, in <module>
  1 Description
    from wellcomeml.ml.bert_vectorizer import BertVectorizer
  File "/Users/nsorros/code/WellcomeML/wellcomeml/ml/bert_vectorizer.py", line 9, in <module>
    import torch
ModuleNotFoundError: No module named 'torch'
```

Checklist
---

- [ ] Added tests

### DIFF
--- a/wellcomeml/ml/__init__.py
+++ b/wellcomeml/ml/__init__.py
@@ -1,3 +1,4 @@
+import traceback
 from wellcomeml.logger import logger
 
 
@@ -33,3 +34,4 @@ try:
 except ImportError as e:
     logger.error(e)
     logger.warning("Using WellcomeML without extras (transformers & torch).")
+    logger.debug(traceback.format_exc())
--- a/README.md
+++ b/README.md
@@ -111,3 +111,8 @@ rustup-init
 
 Examples can be found in the subfolder `examples`.
 
+## 4. Troubleshooting
+
+If you experience a problem with installing or using WellcomeML please open an issue. It might be
+worth setting the logging level to DEBUG `export LOGGING_LEVEL=DEBUG` which will often expose
+more information that might be informative to resolve the issue.


lizgzil:
@nsorros This is v handy, but could do with updating the README too so people know about it?

nsorros:
> @nsorros This is v handy, but could do with updating the README too so people know about it?

eca794d

How about this? Should we say more or something different? For now, it is only one place where we make use of DEBUG but in the future we might add more of those.

wellcomeml/ml/__init__.py
```diff
@@ -33,3 +34,4 @@
 except ImportError as e:
     logger.error(e)
     logger.warning("Using WellcomeML without extras (transformers & torch).")
+    logger.debug(traceback.format_exc())
```
aCampello, line 37: That's great. We should in any case move towards a more granular approach with extras, but having the debug in the meantime is good!

aCampello (approved):
Minus the pipeline problem, this is great! Thanks!

Issue #271, opened 2021-04-07
Investigate spacy lemmatiser in v.3.0


aCampello:
Seem that v3.0 requires tagger, which makes it much slower. Maybe we can load the old look-up one

https://spacy.io/api/lemmatizer

lizgzil:
I think we can replace https://github.com/wellcometrust/WellcomeML/blob/dbb71d863bca32890e703d44c1ffbbb680de760a/wellcomeml/ml/frequency_vectorizer.py#L86-L101

with 
```
nlp = spacy.blank("en")
nlp.add_pipe("lemmatizer", config={"mode": "lookup"})
nlp.initialize()
```
Which on 1000 grant descriptions took 37.5 secs with the current method (v1) and 2.1 secs with the new method (v2).

The performance is different between v1 and v2 when I looked at the output of one grant description - but I think there may actually be a bug since it doesn't look lemmatized in v1:
```
>>> set(v1_first_output).difference(set(v2_first_output))
{'underrepresented', 'included', 'remaining', 'organisations', 'recruited', 'backgrounds', 'greater', 'priorities', 'includes', 'disadvantaged', 'wishes', 'depending', 'years', 'located', 'indicates', 'funds', 'graduates', 'achieving', 'living', 'groups', 'students', 'expenses', 'recommended'}
>>> set(v2_first_output).difference(set(v1_first_output))
{'depend', 'achieve', 'underrepresent', 'recommend', 'university', 'priority', 'include', 'graduate', 'remain', 'organisation', 'fund', 'student', 'indicate', 'locate', 'wish', 'expense', 'great', 'live', 'disadvantage'}
```
Happy to make this change

Issue #270, opened 2021-04-07
Travis takes forever/breaks sometimes


aCampello:
https://travis-ci.com/github/wellcometrust/WellcomeML/jobs/496463155

I think it has to do with the the merge of https://github.com/wellcometrust/WellcomeML/pull/269

PR #269, opened 2021-03-31
Enable tagger which is necessary on V3

aCampello:
Description
---

Fixed #268 .

As of Spacy v3 the lemmatiser needs the tagger!! We were disabling it from the pipeline which made it stop working and gave warnings:
```
[W108] The rule-based lemmatizer did not find POS annotation for the token 'TEs'. Check that your pipeline includes components that assign token.pos, typically 'tagger'+'attribute_ruler' or 'morphologizer'.
2021-03-31 11:04:08 spacy WARNING: [W108] The rule-based lemmatizer did not find POS annotation for the token 'TEs'. Check that your pipeline includes components that assign token.pos, typically 'tagger'+'attribute_ruler' or 'morphologizer'.
```
This fixes it

Checklist
---

- [x] Added link to Github issue or Trello card
- [ ] Added tests

PR #272, opened 2021-04-09
Spacy lemmatizer to use blank model

lizgzil:
Description
---

Fixes https://github.com/wellcometrust/WellcomeML/issues/271 by using blank spacy model.
Running on 1000 grant description: the previous method took 33 secs (when tagger is disabled) and 14 secs (when tagger is not disabled). This new method takes 2 secs.

I also noticed some unusual performance where other methods don't seem to actually lemmatize the text.
```
X = ['the cats sat on the sitting room floors enjoying the sun']
```
**Original method (when the model is already downloaded)**
```
nlp = spacy.load("en_core_web_sm", disable=["ner", "tagger", "parser", "textcat"])
output= [[token.lemma_.lower() for token in doc] for doc in nlp.pipe(X)]
```
gives loads of warning for every token of the form "spacy WARNING: [W108] The rule-based lemmatizer did not find POS annotation for the token 'sun'. Check that your pipeline includes components that assign token.pos, typically 'tagger'+'attribute_ruler' or 'morphologizer'" (allenai seem to have had a [similar issue](https://github.com/allenai/allennlp/issues/5036) with it)
and the output is:
```
[['the', 'cats', 'sat', 'on', 'the', 'sitting', 'room', 'floors', 'enjoying', 'the', 'sun']]
```
I realise this had a bug in it where the tagger shouldn't have been disabled, i.e.
```
nlp = spacy.load("en_core_web_sm", disable=["ner", "parser", "textcat"])
output= [[token.lemma_.lower() for token in doc] for doc in nlp.pipe(X)]
```
does seem to lemmatize correctly.

**I also tried the following method:**
```
from spacy.lang.en import English
from spacy.pipeline.lemmatizer import Lemmatizer
nlp = English()
lemmatizer = Lemmatizer(nlp, model=None,  mode= "lookup")
output= [[lemmatizer.lookup_lemmatize(token)[0].lower() for token in doc] for doc in nlp.pipe(X)]
```
which also gave:
```
[['the', 'cats', 'sat', 'on', 'the', 'sitting', 'room', 'floors', 'enjoying', 'the', 'sun']]
```

**The final method**, which I've commited is:
```
nlp = spacy.blank("en")
nlp.add_pipe("lemmatizer", config={"mode": "lookup"})
nlp.initialize()
output= [[token.lemma_.lower() for token in doc] for doc in nlp.pipe(X)]
```
gives:
```
[['the', 'cat', 'sit', 'on', 'the', 'sit', 'room', 'floor', 'enjoy', 'the', 'sun']]
```

Checklist
---

- [ ] Added tests

### DIFF
--- a/wellcomeml/ml/frequency_vectorizer.py
+++ b/wellcomeml/ml/frequency_vectorizer.py
@@ -83,23 +83,9 @@ class WellcomeTfidf(TfidfVectorizer):
 
         """
 
-        try:
-            nlp = spacy.load(
-                "en_core_web_sm", disable=["ner", "tagger", "parser", "textcat"]
-            )
-        except IOError:
-            from wellcomeml.__main__ import download
-
-            download("models")
-            # pkg_resources need to be reloaded to pick up the newly installed models
-            import pkg_resources
-            import imp
-
-            imp.reload(pkg_resources)
-            nlp = spacy.load(
-                "en_core_web_sm", disable=["ner", "parser", "textcat"]  # Spacy 3.0 needs the tagger to lemmatise
-            )
-
+        nlp = spacy.blank("en")
+        nlp.add_pipe("lemmatizer", config={"mode": "lookup"})
+        nlp.initialize()
 
         logger.info("Using spacy pre-trained lemmatiser.")
         if remove_stopwords_and_punct:


wellcomeml/ml/frequency_vectorizer.py
```diff
@@ -83,23 +83,9 @@ def spacy_lemmatizer(self, X, remove_stopwords_and_punct=True):
...
-            import pkg_resources
-            import imp
-
-            imp.reload(pkg_resources)
-            nlp = spacy.load(
-                "en_core_web_sm", disable=["ner", "parser", "textcat"]  # Spacy 3.0 needs the tagger to lemmatise
-            )
-
+        nlp = spacy.blank("en")
+        nlp.add_pipe("lemmatizer", config={"mode": "lookup"})
```
aCampello, line 87: Nice find. I think this needs `spacy-lookup-data` to be install (which is probably installed automatically with `spacy[lookup]`.

aCampello (approved):
Nice find. Is it much quicker then?

wellcomeml/ml/frequency_vectorizer.py
lizgzil: just did a little experiment with a blank virtualenv and it seems like as long as `spacy[lookups]` is installed then in the script `import spacy` is enough for this to work. Or am I confusing something?

lizgzil:
> Nice find. Is it much quicker then?

Running on 1000 grant descriptions the previous method took around 14 secs (when tagger is not disabled) and this new method takes 2 seconds.

aCampello:
> > Nice find. Is it much quicker then?
> 
> Running on 1000 grant descriptions the previous method took around 14 secs (when tagger is not disabled) and this new method takes 2 seconds.

⚡

aCampello:
Pipeline seems to fail with a weird error.

lizgzil:
> Pipeline seems to fail with a weird error.

@aCampello :( seems to be the same one as you got when you opened https://github.com/wellcometrust/WellcomeML/issues/270. Did you do anything to get your [PR](https://github.com/wellcometrust/WellcomeML/pull/269) to pass in the end? As @nsorros said "Some non deterministic test fail here showing up again."

aCampello:
Don't know, I re-ran and it passed! I assume some memory error again.

aCampello:
Let me have a look. Does `tox` pass locally?

lizgzil:
> Let me have a look. Does `tox` pass locally?

@aCampello it's still running locally for 3.8, but has passed for 3.7

aCampello:
I re-triggered your pipeline. Let's see.

wellcomeml/ml/frequency_vectorizer.py
nsorros: can you try what happens if `en_core_web_sm` model is not installed? I think an IOError is thrown still so you should keep the exception handling that @aCampello had

nsorros (changes requested):
Blocking so we check whether the exception handling is still needed

wellcomeml/ml/frequency_vectorizer.py
lizgzil: seems to work without it being installed:
```
>>> import spacy
>>> nlp=spacy.load('en_core_web_sm')
  File "/Users/gallaghe/Code/nutrition-labels/build/virtualenv/lib/python3.7/site-packages/spacy/__init__.py", line 47, in load
    return util.load_model(name, disable=disable, exclude=exclude, config=config)
  File "/Users/gallaghe/Code/nutrition-labels/build/virtualenv/lib/python3.7/site-packages/spacy/util.py", line 329, in load_model
    raise IOError(Errors.E050.format(name=name))
OSError: [E050] Can't find model 'en_core_web_sm'. It doesn't seem to be a Python package or a valid path to a data directory.
>>> nlp = spacy.blank("en")
>>> nlp.add_pipe("lemmatizer", config={"mode": "lookup"})
<spacy.lang.en.lemmatizer.EnglishLemmatizer object at 0x15ad7ca00>
>>> nlp.initialize()
<thinc.optimizers.Optimizer object at 0x15aba8050>
>>> X = ['the cats sat on the sitting room floors enjoying the sun']
>>> [[token.lemma_.lower() for token in doc] for doc in nlp.pipe(X)]
[['the', 'cat', 'sit', 'on', 'the', 'sit', 'room', 'floor', 'enjoy', 'the', 'sun']]
```

nsorros: ok cool, just realised that you initialise a blank model, good 👍 thanks for checking

nsorros (approved):

wellcomeml/ml/frequency_vectorizer.py
aCampello: Yeah, that's actually a big advantage. You just need to install the lookups, not the entire model! Much simpler code too.

Issue #125, opened 2020-09-02
Improve way to handle extras


aCampello:
At the moment, the way to handle extras is to look whether we can import the functions, and if not, throw a warning.

That might hide other development errors that prevent from import the functions, which is not really related to extras

https://github.com/wellcometrust/WellcomeML/blob/2afcd74a8397b47c10ac1e227168c3ff7dd1f05d/wellcomeml/ml/__init__.py#L46-L47

Ideally we would like to show what is the error besides the warning.

nsorros:
is this resolved by #262? @aCampello

aCampello:
Partially. I think the idea was to have a better handling of extras in general, but that is addressed in other issues, so I think we can close this.

PR #273, opened 2021-04-09
Feature/speed up tests

### LINKED ISSUE #270: Travis takes forever/breaks sometimes

aCampello:
Description
---
Fixed #270 
Addresses #89 

Tweaks some parameters of the test and adds some features to avoid unnecessary reloading models when tests re-use the same model..

Checklist
---

- [ ] Added tests


aCampello:
The results are modest at the moment: reduces timing from 32m to 26m. But that's enough to avoid the error we were having of pipelines failing due to memory.

aCampello:
The expensive tests now are:

```
133.26s call     tests/test_bert_classifier.py::test_scibert
1275127.71s call     tests/test_bert_classifier.py::test_multilabel
1276123.98s call     tests/test_bert_classifier.py::test_save_load
1277122.25s call     tests/test_bert_classifier.py::test_multiclass
127861.83s call     tests/test_bert_vectorizer.py::test_embed_long_sentence
127951.33s call     tests/test_keras_vectorizer.py::test_build_embedding_matrix_word_vectors
128049.82s call     tests/test_bilstm.py::test_save_load_attention
128147.85s call     tests/test_bilstm.py::test_save_load
128242.18s call     tests/test_clustering.py::test_parameter_search[umap]
128319.29s call     tests/test_sent2vec.py::test_fit
128416.92s setup    tests/test_bert_classifier.py::test_multilabel
128514.05s setup    tests/test_bert_vectorizer.py::test_fit_transform_works
128613.03s call     tests/test_bilstm.py::test_multilabel
128712.23s call     tests/test_cnn.py::test_save_load_attention
128812.15s call     tests/test_bilstm.py::test_attention
128912.08s call     tests/test_bilstm.py::test_sparse
129012.02s call     tests/test_bilstm.py::test_early_stopping
129111.88s call     tests/test_bilstm.py::test_predict_proba
129211.87s call     tests/test_bilstm.py::test_vanilla
129311.02s call     tests/test_bilstm.py::test_threshold
129410.09s call     tests/test_cnn.py::test_save_load
12958.44s call     tests/test_clustering.py::test_full_pipeline[umap-True]
12967.31s call     tests/test_bert_vectorizer.py::test_embed_scibert
```

aCampello:
This is WIP but could you have a look @lizgzil @nsorros to see if you have any ideas on how to speed up the tests further?

lizgzil:
some thoughts on `tests/test_bilstm.py`: 
- `test_vanilla`: nb_epochs is at 10, could this be less and perhaps just test that the score is between 0 and 1? I'm not sure if the test actually tests for the model training properly anyway - e.g. it could still be random and give a score >0.6 (less likely obviously).
- `test_multilabel`, `test_sparse`, `test_attention`: similar comment to above (except they use default value of nb_epochs=5)
- `test_predict_proba`, `test_threshold`: these aren't even testing for good training so I think you can set num_epochs = 1.
- `test_save_load` and `test_save_load_attention`: could set nb_epochs=1 (since default is 5). and again `assert loaded_model.score(X_vec, Y) > 0.6` could be a check the score is between 0 and 1, or perhaps don't even have this check?

lizgzil:
also similarly for `test_bert_classifier`, they all use the default `epochs=5`, but this could be 2 since I don't think we are ever testing that the trained models actually perform well.

aCampello:
> also similarly for `test_bert_classifier`, they all use the default `epochs=5`, but this could be 2 since I don't think we are ever testing that the trained models actually perform well.

We actually test if the training loss is decreasing: `assert model.losses[0] > model.losses[-1]`. Weirdly if we only do 2 epochs (or even 3) this is not true. I am not sure if we should be asserting this though. I mean the objective of the test is to make sure the code is functioning, not performance?

lizgzil:
> > also similarly for `test_bert_classifier`, they all use the default `epochs=5`, but this could be 2 since I don't think we are ever testing that the trained models actually perform well.
> 
> We actually test if the training loss is decreasing: `assert model.losses[0] > model.losses[-1]`. Weirdly if we only do 2 epochs (or even 3) this is not true. I am not sure if we should be asserting this though. I mean the objective of the test is to make sure the code is functioning, not performance?

oh yeh, I missed that. Yeh I kind of agree - I think we should just be testing function not performance. The other way of looking at it is that `assert model.losses[0] > model.losses[-1]` (or in the above cases `score>0.6`) could be true by random chance, so the test perhaps isn't useful anyway.

nsorros:
Part of the reason I had not replied so far is that I was wondering whether there is something in the import or loading process that delays all tests or the delay comes purely from running the test. I investigated a bit the bert_classifier tests and it seems that the delay comes indeed from running the tests i.e. fit is slow which makes sense, it is a big model.

The suggestions that liz made about reducing number of epochs makes sense, I guess we can reduce something that still makes sense like 2 or 3 but in most cases we do not need 5 or 10. In cases where we do not need it we should reduce.

What I disagree with is removing the test for decreasing loss or that a score is better than random. I think in both cases the point is that this cannot happen by chance. If a test does not beat chance let's remove it but I think it does at the moment and we want that consistency. That said, if there is a better way to ensure fit is working correctly we are open to suggestions.

Finally what I would want us to investigate a bit more is if there is something that delays the tests that we can control i.e. I have noticed that importing tensorflow and as a result any model takes a lot of time, is there a way to speed this up? Or some slow things might be happening twice in some tests, can we refactor and run once? Or we might want to test that the slow fit is tested once not a lot of times in a test.

nsorros:
Note that `time python -c "import wellcomeml.ml"` this takes 30 seconds and this runs every time we run a test file, not every test but test_file so improving our imports will should have an impact on tests. Or improving where we import, if we can import wellcomeml.ml once this will also improve

nsorros:
The other thing we can do to speed tests up are to mock the calls to tensorflow and transformers and only test our code assuming theirs works. We can possibly test that both work together once a kind of integration test and use mocks everywhere else.

nsorros:
Also this test `tests/test_keras_vectorizer.py::test_build_embedding_matrix_word_vector` uses `glove-twitter-25` which might be a big embedding matrix, is it possible to replace with something smaller? even local?

nsorros:
One more thought, are we running the tests in parallel utilising all cores of the machine Travis is using? In GitHub Actions you can set up to 5 works to run your tests I think.

aCampello:
Those are all valid point. I will experiment with making direct imports to the tests to see how much it speeds it up! This way we can keep the epochs to 5 and still test for some training loss decrease.

aCampello:
It didn't seem to speed up the pipeline.

Did I miss something @nsorros  in https://github.com/wellcometrust/WellcomeML/pull/273/commits/db81e27e09463241446645e64e749bdd30766e06 ?

You mentioned "deleting" __init__ ?

aCampello:
I see, I have to _literally_ delete `__init__.py`. Will try it now.

tests/test_bert_vectorizer.py
```diff
@@ -12,43 +12,50 @@
 ]
 
 
-def test_embed_one_sentence():
+@pytest.fixture
```
nsorros, line 15: I think it will make tests even quicker to invoke once by using scope module https://docs.pytest.org/en/latest/how-to/fixtures.html#fixture-scopes

tests/test_bert_classifier.py
```diff
@@ -1,12 +1,21 @@
...
 import tempfile
 
 import numpy as np
 
 from wellcomeml.ml.bert_classifier import BertClassifier
 
 
-def test_multilabel():
+@pytest.fixture
+def multilabel_bert():
```
nsorros, line 11: we can use scope module here to invoke once https://docs.pytest.org/en/latest/how-to/fixtures.html#fixture-scopes

nsorros:
Btw this is a breaking change so we need to release and mention it. If we want to be true to SemVer we need to upgrade major, if we don't we at the very least need to mention in the changelog

aCampello:
(Agree with the breaking change and calling it 2.0) 

I'm more and more convinced the a good way to do it is by breaking the test environment into different pieces (https://docs.travis-ci.com/user/speeding-up-the-build/). For this we would fix the way to improve extras, #125 then mark the tests as, say "spacy", "bert", "keras"  and "others" and run 4 concurrent travis pipelines. We would still have the overhead of building the venv, but it would parallelise all the rest.

wellcomeml/ml/__init__.py
```diff
@@ -1,37 +1,37 @@
...
+#         'BertClassifier', 'BertVectorizer', 'SpacyKnowledgeBase',
+#         'SpacyEntityLinker', 'SemanticEquivalenceClassifier',
+#         'CNNClassifier', 'BiLSTMClassifier', 'KerasVectorizer',
+#         'SimilarityEntityLinker', 'SemanticEquivalenceClassifier',
+#         'TransformersTokenizer'
+#     ]
+# except ImportError as e:
+#     logger.error(e)
+#     logger.warning("Using WellcomeML without extras (transformers & torch).")
+#     logger.debug(traceback.format_exc())
```
nsorros, line 37: need to remove before merge

wellcomeml/ml/bert_classifier.py
```diff
@@ -114,7 +116,8 @@ def fit(self, X, Y):
            Y: list or numpy array of classes (n_samples, n_classes)
         """
         num_labels = len(Y[0])
-        self._init_model(num_labels)
+        if not self.initiated_:
```
nsorros, line 119: good addition, you could also check for existence of model and tokenizer but this works as well

nsorros (commented):
The changes look good, note that this is a breaking change so we probably want to release mentioning this.

I am waiting for tests to pass to accept

tests/test_bert_classifier.py
aCampello: Good shout, forgot about the scope!

aCampello: https://github.com/wellcometrust/WellcomeML/pull/273/commits/8a9b45a47d03a244e8280180bd4cbb2a4fb06f34

wellcomeml/ml/__init__.py
aCampello: I am still not 100% sure I'll remove this in this PR. Maybe I will put it back and remove it when we address extras in a better way.

aCampello:
WIN!

aCampello:
Splitting the tasks into "Bert" and "non bert" tasks reduces it to 16m (from 32m). I believe we can even scale up and make more sensible split of tests, but I am calling this a victory for now an merging this PR if you guys agree

aCampello:
Can I get you sign off @nsorros @lizgzil

nsorros (approved):
This is a great impovement but also ⚠️ a breaking change. I suggest you release a new version and record the breaking API change

tox.ini
```diff
@@ -7,5 +7,5 @@ deps =
 	.[deep-learning]
 
 commands = python -m spacy download en_core_web_sm
-           pytest -m  "not (integration)" -s -v --durations=0 --disable-warnings --tb=line --cov=wellcomeml ./tests
+           pytest -m '{env:TEST_SUITE:}' -s -v --durations=0 --disable-warnings --tb=line --cov=wellcomeml ./tests
```
nsorros, line 10: is there a default value? what happens locally in the absence of this variable?

aCampello: In the absence of the value it just throws an empty string and runs all test, which is probably what we want locally.

aCampello:
> This is a great impovement but also ⚠️ a breaking change. I suggest you release a new version and record the breaking API change

Hey Nick, I put back the `__init__.py` file, so I don't think it's a braking change anymore (you can still import the old way). Or was there anything else that breaks the previous version?

nsorros:
> > This is a great impovement but also ⚠️ a breaking change. I suggest you release a new version and record the breaking API change
> 
> Hey Nick, I put back the `__init__.py` file, so I don't think it's a braking change anymore (you can still import the old way). Or was there anything else that breaks the previous version?

No that's what I meant

Issue #140, opened 2020-09-18
TextClustering additions


lizgzil:
Some extra functionality for the TextClustering class would be useful.

- [x] Include random seeds for reproducibility
- [x] Separate out reduction and clustering functionalities - sometimes we just might need to reduce, so a shame to spend ages waiting for clustering too
- [x] Expose keywords dict `kw_dictionary`
- [x] Calculate and expose cluster mid point dict {cluster_num: coord,...} - this can be useful to plot the keywords for each cluster
- [x] Save and load functions - depending on the use case you might want to save the embedded points so you don't need to re-run

PR #276, opened 2021-04-16
Feature/clustering improvements

aCampello:
Description
---

Closes https://github.com/wellcometrust/WellcomeML/issues/140

- Separate out reduction and clustering functionalities - sometimes we just might need to reduce, so a shame to spend ages waiting for clustering too
- Save and load functions - depending on the use case you might want to save the embedded points so you don't need to re-run

Checklist
---

- [x] Added tests

### DIFF
--- a/wellcomeml/ml/clustering.py
+++ b/wellcomeml/ml/clustering.py
@@ -249,7 +249,6 @@ class TextClustering(object):
                param_grid.get('clustering', {}).items()}
         }
 
-
         grid = GridSearchCV(
             estimator=pipeline,
             param_grid=params,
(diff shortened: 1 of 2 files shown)


aCampello:
The pipeline won't pass yet because of the other PR #273, but the clustering tests pass locally.

aCampello:
Don't know why codecov wasn't triggered, but basically this will increase coverage by 1%-ish https://app.codecov.io/gh/wellcometrust/WellcomeML/compare/276/overview

wellcomeml/ml/clustering.py
```diff
@@ -292,6 +304,74 @@ def optimise(self, X, param_grid, n_cluster_range=None, max_noise=0.2,
...
+            folder(str): path to folder
+            components(list or 'all'): List of components to load. Options are: 'embbedded_points',
+            'reduced_points', 'vectorizer', 'reducer', and 'clustering_model'. By default, loads
+            'all' (you can get all components by listing the class param
+             TextClustering.components)
+
+        """
+
+        if components == 'all' or 'embedded_points' in components:
+            self.embedded_points = np.load(os.path.join(folder, 'embedded_points.npy'),
```
nsorros, line 354: there is some duplication about saving and loading the paths. you can create them in a variable that stays in the class and reuse `self.embedded_points_path = xxx` possibly in init or something

nsorros (approved):

wellcomeml/ml/clustering.py
aCampello: As I pushed the code I thought "Is Nick going to comment on code duplication?" 😆 . 

That's a very fair point, will address.

PR #279, opened 2021-04-27
Extend logger to more external libraries

### LINKED ISSUE #280: Propagate LOGGING LEVEL to external libraries and add more external libraries
Currently the only library that we set the level for is `transformers` where we set the level to WARNING irrespective of the `LOGGING_LEVEL` that a user might have set. The proposal is to to use the `LOGGING_LEVEL` as a global variable that controls logging in all libraries that WellcomeML uses e.g. tensorflow, gensim, spacy, tokenisers and more

nsorros:
Description
---

The main motivation for ths PR has been to extend our environment
variable that controls logging to more external libraries so that
we can control the logging level more globally.

As I was investigating how we log, I noticed that we mainly import
from the logger module which removes the information about the file
that triggers the logger. So I replaced that with 
`logging.getLogger(__name__)` to include that information in the
logging message which I think was our intention. It is also what
is recommended https://docs.python.org/3.5/howto/logging.html#logging-advanced-tutorial

It would also be nice to set a custom logging level per library but I am not 100% sure whether this is needed. Added an issue for it #281 

I still want to resolve some issues that have arisen which are
described below.

Fixes #280 

## Gensim warnings

```
Users/nsorros/code/WellcomeML/build/virtualenv/lib/python3.8/site-packages/gensim/similarities/__init__.py:15: UserWarning: The gensim.similarities.levenshtein submodule is disabled, because the optional Levenshtein package <https://pypi.org/project/python-Levenshtein/> is unavailable. Install Levenhstein (e.g. `pip install python-Levenshtein`) to suppress this warning.
  warnings.warn(msg)
```

**Update**:
It turns out that this warning is created from `warnings` not a logger and the way to disable python warnings is by passing a flag or setting an env variable or using the warnings package to filter. See https://stackoverflow.com/questions/14463277/how-to-disable-python-warnings. Not sure if this is still in scope for this PR.

**Update 2**:
Disabled warning when user sets the logging level to ERROR or higher.

## Duplication in Spacy
```
[2021-04-27 12:01:00,234] [INFO] Added vocab lookups: lexeme_norm
2021-04-27 12:01:00 spacy INFO: Added vocab lookups: lexeme_norm
[2021-04-27 12:01:00,235] [INFO] Created vocabulary
2021-04-27 12:01:00 spacy INFO: Created vocabulary
[2021-04-27 12:01:00,235] [INFO] Finished initializing nlp object
2021-04-27 12:01:00 spacy INFO: Finished initializing nlp object
```

**Update**
That behaviour preexists this PR. Opened an issue #282 

Checklist
---

- [ ] Added tests

### DIFF
--- a/wellcomeml/logger.py
+++ b/wellcomeml/logger.py
@@ -39,7 +39,10 @@ external_logging_level = {
     'transformers': NUMERIC_LEVEL,
     'tensorflow': NUMERIC_LEVEL,
     'gensim': NUMERIC_LEVEL,
-    'sklearn': NUMERIC_LEVEL
+    'sklearn': NUMERIC_LEVEL,
+    'spacy': NUMERIC_LEVEL,
+    'torch': NUMERIC_LEVEL,
+    'tokenizers': NUMERIC_LEVEL
 }
 
 for package, level in external_logging_level.items():
--- a/wellcomeml/logger.py
+++ b/wellcomeml/logger.py
@@ -31,22 +31,22 @@ def build_logger(logging_level, name):
 
 DEFAULT_LOGGING_LEVEL = "INFO"
 LOGGING_LEVEL = os.getenv("LOGGING_LEVEL", DEFAULT_LOGGING_LEVEL)
-NUMERIC_LEVEL = get_numeric_level(LOGGING_LEVEL)
+LOGGING_LEVEL = get_numeric_level(LOGGING_LEVEL)
 
-logger = build_logger(logging_level=NUMERIC_LEVEL, name=__name__)
+logger = build_logger(logging_level=LOGGING_LEVEL, name=__name__)
 
 external_logging_level = {
-    'transformers': NUMERIC_LEVEL,
-    'tensorflow': NUMERIC_LEVEL,
-    'gensim': NUMERIC_LEVEL,
-    'sklearn': NUMERIC_LEVEL,
-    'spacy': NUMERIC_LEVEL,
-    'torch': NUMERIC_LEVEL,
-    'tokenizers': NUMERIC_LEVEL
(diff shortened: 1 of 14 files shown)


nsorros:
[Control loggers in WellcomeML with env variable.](https://trello.com/c/pimV4FpX/1232-control-loggers-in-wellcomeml-with-env-variable)

wellcomeml/logger.py
```diff
@@ -27,11 +29,24 @@ def build_logger(logging_level, name):
     return logger
 
 
-LOGGING_LEVEL = os.getenv("LOGGING_LEVEL")
+DEFAULT_LOGGING_LEVEL = "INFO"
+LOGGING_LEVEL = os.getenv("LOGGING_LEVEL", DEFAULT_LOGGING_LEVEL)
+NUMERIC_LEVEL = get_numeric_level(LOGGING_LEVEL)
 
-logger = build_logger(logging_level=LOGGING_LEVEL, name=__name__)
+logger = build_logger(logging_level=NUMERIC_LEVEL, name=__name__)
```
aCampello, line 36: A bit clunky terminology. I would maybe use a union type `[str, int]` for LOGGING_LEVEL and re-use it instead of numeric_level?

aCampello (approved):
It seems to me its "INFO" by default, which is probably the right level 👍 . I was a bit concerned about travis pipelines, but I think it should still be the same there.

wellcomeml/logger.py
nsorros: So you suggest we convert `LOGGING_LEVEL` and reassign. Sounds good 👍

nsorros: 7d9ad96fdeda86b0cfe4189ebba89646b328821d

PR #283, opened 2021-04-28
Import umap during init and if needed

nsorros:
Description
---

Apparently what the source of the slowdown of `wellcomeml.ml`
is umap, see screenshot below and https://github.com/lmcinnes/umap/issues/631

As such this PR moves the import inside clustering init and
only when the user selects to use umap. This improves import time
from ~30s to ~10s in my laptop.

Screenshot 2021-04-28 at 11 43 22 AM
Note that `wellcomeml.ml.frequency_vectorizer` seems to take time but I think this is an error. In case of interest the tool used is https://github.com/nschloe/tuna

Checklist
---

- [ ] Added tests


aCampello:
Exciting times

aCampello (approved):

nsorros:
[Make WellcomeML init faster.](https://trello.com/c/4XIeEMOS/1233-make-wellcomeml-init-faster)

PR #302, opened 2021-06-11
Add a Windows specific Makefile and some instructions

SamDepardieu:
Description
---
This PR aims to make the setup process compatible with Windows machines. To this purpose:
 - Add a Makefile.win for Windows users to run
 - Add a WINDOWS_USERS.md readme file with instructions

Note: Due to some issues with the installation of sent2vec, the windows Makefile will pull the repository and install it directly from the local clone, hence the addition of sent2vec into the gitignore file.

Checklist
---

- [ ] Added tests


nsorros:
@SamDepardieu If the main difference is the `VIRTUALENV` and `PYTHON` location why don't we check at the begging of the file the OS of the user and initialise according so that we can have one Makefile?

SamDepardieu:
@nsorros You do certainly get a point. Will edit this to add a system check on the main Makefile instead :+1:

nsorros (commented):
Looks good but this is still a separate Makefile, can we not have one Makefile given that you have abstracted the differences between Windows and Unix like OS

SamDepardieu:
@nsorros I was just editing the `.win` makefile to pull it on my Windows machine for testing purpose, pushed the actual changes :)

nsorros (approved):
Great addition 💪 I guess this is out of scope but if we could add a test that tests run successfully on a Windows container that would be user

Makefile
```diff
@@ -28,15 +36,18 @@ dist: update-docs
...
 
 $(VIRTUALENV)/.deep_learning_models:
-#	$(VIRTUALENV)/bin/python -m spacy download en_trf_bertbaseuncased_lg
+#	$(VENV_BIN)/python -m spacy download en_trf_bertbaseuncased_lg
 	touch $@
 
 $(VIRTUALENV)/.non_pypi_packages:
-	$(VIRTUALENV)/bin/pip install git+https://github.com/epfml/sent2vec.git
+	# Install from local git directory - pip install [address] fails on Windows
+	git clone https://github.com/epfml/sent2vec.git
```
aCampello, line 48: Nightmare! :S I am not sure how much we use sent2vec at the moment, maybe we should drop it. @nsorros ?

WINDOWS_USERS.md
```diff
@@ -0,0 +1,19 @@
+# Installation on Windows machines
```
aCampello, line 1: ⭐

aCampello (commented):
🌟 ! Just waiting on one of the test pipelines to pass on travis, then you can merge.

SamDepardieu:
Sadly I'm fairly sure this wouldn't run great in a container... It hardly runs on a physical machine as it needs some build specific libs that Jeff couldn't even get working on Vagrant. If you just want us to test for OS specific routes though, it's mostly about testing is the `OS` env variable is set (it is by default on all Windows systems), so that's something fairly easy to test for

jdu:
I think in the case of Travis/CodeBuild the build environment is actually a virtualized WIndows Server instance which is similar to Windows 10 in most ways, however, we may have to find ways to install the required dependencies for building on windows in an automated way before this is diable.

The issues I had with vagrant were around VirtualBox I believe, effectively the only way around that is to install tensorflow-cpu instead of tensorflow because the Vagrant/VirtualBox VM has no access to the host systems GPU, and ends up installing/compiling for the FMA instruction set which then doesn't exist in the VM. That's an issue fairly specific to VirtualBox/Vagrant though and may get resolved by the 3.0 rollout of Vagrant.

So it's possible it may well work under CodeBuild / Travis for windows builds, just we may have to look at some additional scripts / tooling to run on windows builds before the `Makefile` to ensure that the pre-requisites are installed like MSVC++, Cygwin, etc... as I don't think Windows Server has the Linux subsystem like on Windows 10.

aCampello (approved):
We tested it locally and it works too. About adding a runner and other tests etc, we can potentially address as #258 or open other issues if we find pertinent.

Issue #284, opened 2021-04-28
Delete wellcomeml.ml init?


nsorros:
We have been discussing whether we want to continue having an `__init__` function that exposes all the Wellcomeml ml models. This is convenient but running any model as slow as the slowest model. This was more of a problem before #283 where imports used to take ~30s but currently this has been reduced to ~12s which borderline acceptable.

Just to given an indication of the different speeds of imports in case `__init__` was deleted
* vectorizer ~10s
* bert_vectorizer ~3s
* bert_classifier ~9s
* cnn ~ 8s

What seems to delay the imports is tensorflow which is widely used and alone it takes 7s to import.Note that on the surface it seems that spacy takes 10s to load but this is only because in the presence of tensorflow and torch, thinc (the underlying ml library of spacy) makes api connections to them so it imports them.

As it is clear we would probably not be discussing deleting `wellcomeml.ml` init if it was not for the slowdown. Further to that it seems difficult, but maybe possible, to do further improvements on the speed as imports are interconnected. For example importing spacy when used does not improve the speed as even though it removes the slowdown due to the connection to tensorflow and torch, these libraries will be imported elsewhere anyway so the overall time remains the same. Tensorflow which seems to be the library slowing down the imports is so widely used that it would be difficult to import only when used, this would change a lot of files but it is an avenue we could explore.

aCampello:
This is definitely a pain. I'm currently using `WellcomeTfIdf` in a project and getting a couple of seconds of buffer every time, even though I am not using tensorflow at all. Solving the extras #286 should help too.

aCampello:
This will be partially addressed by https://github.com/wellcometrust/WellcomeML/pull/306, but not quite. Users will be able to disable it with an env variable.

PR #306, opened 2021-06-23
DS-2 Feature/extras

### LINKED ISSUE #234: Is Deep Learning really an extra?
I think in the beginning we were thinking of using wellcomeml for IO operations as well, however I myself never `pip install wellcomeml` without extras. I wonder if it is time to promote deep learning as the core library, and leave the extras for any real extras we might want to add.

### LINKED ISSUE #285: Introduce additional extras to provide better control of which libraries are installed
This is related to #234 and to a certain extent assumes that `deep-learning` will be deprecated in favour of more refined extras.

We need to discuss our motivation for doing so to frame the discussion. A lot of libraries provide a level of control about which libraries are installed with extras for example spacy (spacy[lookups, cuda,...]), dvc (dvc[s3]). I think the rationale in most cases is that there is a core functionality that most users need and some additional functionality that is left as extra to reduce the size of the library and possibly import time as this is not always needed.

I think in our case our main concern has always been the size of WellcomeML due to the heavy deep learning libraries such as tensorflow and torch. As such possibly the size of the libraries and where and how they are used is a good farming of the conversation. For example if we are to remove tensorflow, what functionality is left and to which extent it is a core library we depend on. 

We can inspect the size of the library (WellcomeML) and individual dependencies through the virtualenv that it creates, even though few of those dependencies are not always installed like the test dependencies. The size of the virtualenv is ~2GB which makes WellcomeML quite heavy. The biggest libraries we depend on are
* tensorflow 760MB (with some additional that take more space e.g. tensorboard 30M)
* torch 388MB
* spacy 75MB
* pandas 51MB
* sklearn 31MB
* transformers  15MB
* scipy 126MB
This is not a completely accurate image as each of these libraries have further dependencies but it gives us an estimate.

Unsurprisingly the biggest problem is caused by tensorflow and torch followed by spacy and in particular tensorflow. Unfortunately tensorflow is being used far more than torch. Personally I think libraries below 100MB are not a problem at this point including scipy.

**tensorflow**
```
wellcomeml/ml/attention.py:import tensorflow as tf
wellcomeml/ml/bert_classifier.py:import tensorflow as tf
wellcomeml/ml/bert_semantic_equivalence.py:import tensorflow as tf
wellcomeml/ml/bilstm.py:import tensorflow as tf
wellcomeml/ml/cnn.py:import tensorflow_addons as tfa
wellcomeml/ml/cnn.py:import tensorflow as tf
wellcomeml/ml/keras_utils.py:import tensorflow as tf
```

**torch**
```
wellcomeml/ml/bert_vectorizer.py:import torch
wellcomeml/ml/spacy_classifier.py:import torch
```

**spacy**
```
wellcomeml/ml/frequency_vectorizer.py:#import spacy
wellcomeml/ml/spacy_classifier.py:import spacy
wellcomeml/ml/spacy_entity_linking.py:import spacy
wellcomeml/ml/spacy_knowledge_base.py:import spacy
wellcomeml/ml/spacy_ner.py:import spacy
```

## Suggestions
Most of our code at the moment depends on tensorflow so I am not sure we can eliminate that, we can definitely make it an explicit extra though and throw an error if the user tries to import a library that needs it. But is it an extra?

We can do the same for spacy. Spacy I think is definitely an extra library for WellcomeML (personal opinion @lizgzil may feel different). In any case if we make tensorflow an extra, I do not see why we should not make spacy also.

Lastly torch is only used in two places and I feel like it can be eliminated. Bert vectorizer can switch to using tensorflow or at least import torch only if needed (maybe scibert is only torch). Then spacy classifier uses torch for GPU only so again we can import if needed or investigate using tensorflow if possible.

Note that transformers does not install tensorflow or torch even though it needs one of them to run. So a completely different avenue is to leave that to the user and just throw a warning that a library is needed. Not sure if this is the right direction.

### LINKED ISSUE #290: Create a list of classifiers/algorithms and put it in a table in the documentation
As per description.

Would be good to have a table that can go in the docs saying: `Algorithm | Type (supervised, unsupervised, etc) | Path (e.g. `wellcomeml.ml.WellcomeTfIdfVectorizer`

aCampello:
Description
---

Fixes #285 
Fixes #234 
Fixes #290 

Addresses #284 

This PR breaks extras into `spacy`, `tensorflow` and `torch`. This way, users that only need only one of those frameworks (or non, if they are using say clustering/voting classifier) will not need to install heavy dependencies.

I also adds a fair amount of documentation around what classes we have so far and dependencies (incidentally this fixes #290 )

**[RFC]** As a proposal, I did not completely delete `wellcomeml.ml.init`. Instead, I created an env variable `DISABLE_DIRECT_IMPORTS` that will essentially toggle it off. If it is toggled on, it raises a FutureWarning saying it importing directly will be removed. Alternatively we can really just remove it completely, and call ourselves 2.0.0. I'd be open to suggestions.

Checklist
---

- [ ] Added tests


tests/test_extras.py
```diff
@@ -0,0 +1,47 @@
...
+]
+
+
+def extra_import_check(module_name, extra_name):
+    with mock.patch.dict(sys.modules, {extra_name: None}):
+        import_module(module_name)
+
+
+@pytest.mark.parametrize("module_name,extra_name", module_extra_pairs)
+def test_dependencies(module_name, extra_name):
```
nsorros, line 43: nice set

nsorros:
That is a massive undertaking, it looks good so far. No comments. Let's get this to the finish line 💪

aCampello:
All right, I believe this is ready for review. Ignore the `html` files (this is just documentation) when reviewing.

aCampello:
I updated the description too with the rationale!

wellcomeml/ml/vectorizer.py
```diff
@@ -30,17 +26,26 @@ def __init__(self, embedding="tf-idf", cache_transformed=False, **kwargs):
...
         }
 
         if not vectorizer_dispatcher.get(embedding):
             raise ValueError(f"Model {embedding} not available")
 
-        self.vectorizer = vectorizer_dispatcher.get(embedding)(**kwargs)
+        vectorizer_path = '.'.join(vectorizer_dispatcher.get(embedding).split('.')[:-1])
+        vectorizer_class_path = vectorizer_dispatcher.get(embedding).split('.')[-1]
+        vectorizer = getattr(
+            __import__(vectorizer_path, fromlist=[vectorizer_class_path]),
```
nsorros, line 44: should't you try to import first and throw an error in case user does not have the library needed?

nsorros (approved):
LGTM 👍 Legend effort 🥇 Just a minor comment in the `vectorizer`

wellcomeml/ml/vectorizer.py
aCampello: Yes, that's what will happen because those modules have now the `throw_extra_import_messages`.

What I wanted to avoid, in addition, was to import all the modules unnecessarily. Say I have everything installed but just want to use tf-idf - then it just imports it.

nsorros: 👍

PR #307, opened 2021-07-02
Create plot.py

federikovi:
Description
---

Checklist
---

- [ ] Added tests


wellcomeml/viz/visualize_clusters.py
```diff
@@ -0,0 +1,31 @@
+from bokeh.plotting import figure, show
+from wellcomeml.viz.palettes import Wellcome33
+from bokeh.io import output_notebook
+output_notebook()
```
aCampello, line 4: I'd actually not call `output_notebook` here, as users might actually want the `html` file instead.

wellcomeml/viz/visualize_clusters.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+def visualize_clusters(reduced_points: list):
+
```
aCampello, line 8: Remove extra line

wellcomeml/viz/visualize_clusters.py
```diff
@@ -0,0 +1,31 @@
...
+    """
+    Visualises clusters and shows basic information
+    Args:
+        reduced_points: list
+        List of list of reduced points. Available at cluster.reduced_points
+
+    Returns:
+        p
```
aCampello, line 16: It actually returns None. You can just say: Returns: None (prints a bokeh figure to file or to notebook)

wellcomeml/viz/visualize_clusters.py
```diff
@@ -0,0 +1,31 @@
...
+
+    """
+
+    blue_wellcome = str(Wellcome33[0])
```
aCampello, line 20: Nice!

wellcomeml/viz/visualize_clusters.py
```diff
@@ -0,0 +1,31 @@
+from bokeh.plotting import figure, show
+from wellcomeml.viz.palettes import Wellcome33
+from bokeh.io import output_notebook
```
aCampello, line 3: Import `output_file` too (see below for a suggestion)

wellcomeml/viz/visualize_clusters.py
```diff
@@ -0,0 +1,31 @@
...
+output_notebook()
+
+
+def visualize_clusters(reduced_points: list):
```
aCampello, line 7: Let's make it customisable! Maybe you can add `radius` and `alpha` as parameters. I'd also include the following parameters

`radius: float, alpha: float, output_file_path: str, output_notebook: bool`

If someone says `output_notebook=True`, then you call output_notebook, if they pass a file_name, then you call `output_file`.

aCampello (changes requested):
Thanks for this PR - it's the start of the dashboard :). Some suggestions to make it more customisable and make our lives easier in the future.

wellcomeml/viz/visualize_clusters.py
```diff
@@ -0,0 +1,31 @@
...
+
+    blue_wellcome = str(Wellcome33[0])
+    TOOLS = ('hover, crosshair, pan, wheel_zoom, zoom_in, zoom_out,'
+             'box_zoom, undo, redo, reset, tap, save, box_select,'
+             'poly_select, lasso_select,')
+    TOOLTIPS = [("index", "$index"), ("(x,y)", "($x, $y)"), ]
+
+    p = figure(tools=TOOLS, tooltips=TOOLTIPS)
+
+    p.scatter(reduced_points[:, 0], reduced_points[:, 1], radius=0.30,
```
aCampello, line 28: Can we add a background= `WellcomeBackground`  ? `from wellcomeml.viz.palettes`

wellcomeml/viz/visualize_clusters.py
federikovi: Changed to html

wellcomeml/viz/visualize_clusters.py
federikovi: 👍  done

wellcomeml/viz/visualize_clusters.py
federikovi: Done

wellcomeml/viz/visualize_clusters.py
federikovi: Sure! I changed it and the result is really nice

wellcomeml/viz/visualize_clusters.py
federikovi: @aCampello flake8 actually requires this extra line

wellcomeml/viz/visualize_clusters.py
federikovi: I have added `radius: float, alpha: float` and `output_notebook: bool` that prints in the notebook if True, on a new page otherwise.
I still haven't added `output_file_path: str` because I don't understand what the string would be. The file path or the file name?

Thanks for the comment!

aCampello: Thanks for the additions. So at the moment you call `ouput_file('cluster_viz.html')`, that's hard coded. with output_file_path you would let the user specify it (maybe have a default). so:

`output_file_path: str = 'cluster_viz.html'`

Btw, can we make all variable lower cased? :) That's usually the convention we follow with python variables inside functions

federikovi: True, it's hard coded so it's better to change it. I also forgot to add in the comment a short explanation of the new parameters.

Regarding lower case it's fine, i just used upper case for not having overriding memory :)

aCampello:
It all looks good to me now! Thanks a lot. If you want to add the new parameter and lower case the arguments, you're free to merge it 🎉🎉🎉

aCampello:
Let's merge this!

aCampello (approved):

PR #308, opened 2021-07-05
Add EPMC client

nsorros:
Description
---

Adds an EPMC client, previously in the datalabs private repository

Checklist
---

- [x] Added tests

### DIFF
--- a/tests/io/epmc/test_client.py
+++ b/tests/io/epmc/test_client.py
@@ -2,30 +2,33 @@ import pytest
 
 from wellcomeml.io.epmc.client import EPMCClient
 
+
 @pytest.fixture
 def epmc_client():
     return EPMCClient(
         max_retries=3
     )
 
+
 def test_search(epmc_client):
     pmid = '24889800'
     query = "ext_id:%s" % pmid
     session = epmc_client.requests_session()
     search_result = epmc_client.search(session, query)
-    assert search_result['pmid']==pmid
+    assert search_result['pmid'] == pmid
+
 
 def test_get_references(epmc_client):
     pub_id = '22749442'
     source = 'MED'
     session = epmc_client.requests_session()
     references = epmc_client.get_references(session, pub_id, source=source)
-    assert len(references)!=0
+    assert len(references) != 0
+
 
 def test_get_citations(epmc_client):
     pub_id = '22749442'
     source = 'MED'
     session = epmc_client.requests_session()
     citations = epmc_client.get_citations(session, pub_id, source=source)
-    assert len(citations)!=0
-    
+    assert len(citations) != 0
--- a/tests/io/epmc/test_client.py
+++ b/tests/io/epmc/test_client.py
@@ -1,3 +1,4 @@
+from unittest.mock import MagicMock
 import pytest
 
 from wellcomeml.io.epmc.client import EPMCClient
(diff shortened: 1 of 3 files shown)


aCampello:
Code coverage got a bit crazy

aCampello:
I'm not sure why the CODECOV of the PR is showing weird stats, but it seems correct in the codecov source: https://app.codecov.io/gh/wellcometrust/WellcomeML/compare/308/tree. I suspect it was some sync issue.

wellcomeml/io/epmc/client.py
```diff
@@ -0,0 +1,302 @@
...
+                )
+                go_to_next_page = False
+
+        if citations:
+            return citations
+        else:
+            return None
+
+
+if __name__ == "__main__":
```
aCampello, line 261: I'm not sure if this script belongs here. Maybe as an 'example' inside `wellcomeml/examples`, or maybe as part of a CLI, as you mentioned? The way it stands, people that only install the library from pip won't be able to use `__main__`.

wellcomeml/io/epmc/client.py
```diff
@@ -0,0 +1,302 @@
...
+            session: requests.Session
+            dois: list of DOI strings
+
+        Yields:
+            Dicts of EPMC responses
+        """
+        query = '"' + '" OR "'.join(dois) + '"'
+        return self.search(session=session, query=query, only_first=False, **kwargs)
+
+    def get_full_text(self, session, pmid):
```
aCampello, line 154: This function is not tested - would be good to add a test, as I suspect it's a tricky one.

wellcomeml/io/epmc/client.py
```diff
@@ -0,0 +1,302 @@
...
+            "pageSize": page_size
+        }
+
+        return self._execute_query(session, params, only_first, **kwargs)
+
+    def search_by_doi(self, session, doi, **kwargs):
+        query = "doi:%s" % doi
+        return self.search(session, query, only_first=True, **kwargs)
+
+    def search_by_pmid(self, session, pmid, **kwargs):
```
aCampello, line 124: `search_by_pmid`, `search_by_pmids` and `search_by_doi`, `search_by_dois` are not tested either (mea culpa!). If you want to add tests for those, that'd be really nice, otherwise, let's open a ticket and work on it later.

aCampello (changes requested):
I don't think the `__main__` should live inside `epmc/client.py`. Unless we have a CLI, this should probably go to examples?

wellcomeml/io/epmc/client.py
nsorros: It does not even work I think. I will put something in examples

nsorros:
> I'm not sure why the CODECOV of the PR is showing weird stats, but it seems correct in the codecov source: https://app.codecov.io/gh/wellcometrust/WellcomeML/compare/308/tree. I suspect it was some sync issue.

I think we may have had that before? Let's investigate if this persists

wellcomeml/io/epmc/client.py
nsorros: all of those are pretty thin wrappers around search. they are quite easy to test, I will just patch search and make sure it is called with the right args

wellcomeml/io/epmc/client.py
nsorros: Same as the others, just need to ensure search is called with the right args. This assumes search is tested. Actually I noticed that the tests call the API which needs to change as well.

aCampello: I think we do want to test calling the API? As in, if they modify their response, we would like to know that it is broken. Probably mark the test as `pytest.mark.integrations`

wellcomeml/io/epmc/client.py
nsorros: 7c3b2293a6fa061eb878e5b78debd81b27b51a6f

wellcomeml/io/epmc/client.py
nsorros: 7c3b2293a6fa061eb878e5b78debd81b27b51a6f

aCampello:
Should I re-review?

nsorros:
> Should I re-review?

Go for it

tests/io/epmc/test_client.py
```diff
@@ -0,0 +1,88 @@
...
+
+@pytest.fixture
+def epmc_client():
+    return EPMCClient(
+        max_retries=3
+    )
+
+
+def test_search(epmc_client):
+    epmc_client._execute_query = MagicMock()
```
aCampello, line 15: Magic ✨

aCampello (approved):
New additions sound good (I didn't;t review the client itself that thoroughly because, well, was already reviewed/working in `datalabs`) 

Thanks a lot for this! I am happy that WellcomeML is now home of this functionality

PR #313, opened 2021-07-14
Fix concat feature approach in CNN

### LINKED ISSUE #314: Feature approach concat does not work
Currently this approach has been implemented with `tf.concat` which expects a list of tensors to concatenate. Instead CNN constructs a matrix of attention_heads, embeddeding_dimension so the concat essentially needs to just flatten that into one vector

nsorros:
Description
---

As I was trying to experiment with multiple attention heads and concatenating them to increase the representational power of the network I released that the implementation was not working as expected. 

Fixes #314 

Checklist
---

- [x] Added tests

### DIFF
--- a/wellcomeml/ml/cnn.py
+++ b/wellcomeml/ml/cnn.py
@@ -224,7 +224,7 @@ class CNNClassifier(BaseEstimator, ClassifierMixin):
         elif self.feature_approach == 'sum':
             x = tf.keras.backend.sum(x, axis=1)
         elif self.feature_approach == 'concat':
-            x = tf.concat(x, axis=1)
+            x = tf.keras.layers.Flatten()(x)
         else:
             raise NotImplementedError
         x = tf.keras.layers.Dense(
--- a/tests/test_cnn.py
+++ b/tests/test_cnn.py
@@ -25,6 +25,23 @@ def test_vanilla():
     assert model.score(X, Y) > 0.6
 
 
+def test_feature_approach_concat():
+    X = [
+        "One",
+        "One only",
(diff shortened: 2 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -19,7 +19,7 @@ jobs:
       - name: Setup Python
         uses: actions/setup-python@v1
         with:
-          python-version: 3.7.10
+          python-version: 3.7.11
```
aCampello, line 22: rebase

wellcomeml/ml/cnn.py
```diff
@@ -224,7 +224,7 @@ def residual_attention(x1):
         elif self.feature_approach == 'sum':
             x = tf.keras.backend.sum(x, axis=1)
         elif self.feature_approach == 'concat':
-            x = tf.concat(x, axis=1)
+            x = tf.keras.layers.Flatten()(x)
```
aCampello, line 227: So you wanted one big one-dimensional vector and that one was only squashing one dimension?

aCampello (approved):

wellcomeml/ml/cnn.py
nsorros: the one there was trying to concatenate two matrix on a particular axis instead of squashing so for example if you had 2 matrices of `(32,25,400)` it would create `(32,50,400)` whereas I only had one matrix to begin with and I wanted a matrix `(32,10_000)`

.github/workflows/main.yml
nsorros: done

PR #315, opened 2021-07-15
Extend tokenizers to accomodate shap text explainer

### LINKED ISSUE #316: Add decode to tokenizers
Transformers and Keras tokenisers do not have a decode method that received encoded text and turns it into text. This is handy in a couple of cases but one of them is to be able to explain the predictions using SHAP.

### LINKED ISSUE #317: Extend Keras tokenizer to accept string not only list of strings
Currently Keras tokenizer only works for list of strings in `encode`, it would be nice to harmonise it with transformers tokeniser by adding the ability to encode a single string. This is also a prerequisite for using SHAP to explain the predictions

nsorros:
Description
---

In order to use the text masker in SHAP a tokenizer needs to implement a certain API, see https://github.com/slundberg/shap/pull/2084

Our tokenisers mostly follow that API (as we created both the API and the tokenisers :p) but some changes are still required to be able to use them with SHAP. One missing method is the ability to decode a single or multiple encoded texts. Another is to accept numpy arrays and work on both lists and single text instances a feature that transformers tokenizer has but keras tokeniser does not have

Fixes #317
Fixes #316 

Checklist
---

- [x] Added tests

### DIFF
--- a/wellcomeml/ml/transformers_tokenizer.py
+++ b/wellcomeml/ml/transformers_tokenizer.py
@@ -109,6 +109,8 @@ class TransformersTokenizer:
             raise NotImplementedError
 
     def decode(self, encoded_text):
+        if not encoded_text:
+            return ""
         if type(encoded_text[0]) in [list, np.ndarray]:
             return self.tokenizer.decode_batch(encoded_text)
         return self.tokenizer.decode(encoded_text)
--- a/wellcomeml/ml/transformers_tokenizer.py
+++ b/wellcomeml/ml/transformers_tokenizer.py
@@ -109,7 +109,8 @@ class TransformersTokenizer:
             raise NotImplementedError
 
     def decode(self, encoded_text):
-        # Note that we currently only support decoding a single encoded sentence
+        if type(encoded_text[0]) in [list, np.ndarray]:
+            return self.tokenizer.decode_batch(encoded_text)
         return self.tokenizer.decode(encoded_text)
 
     def save(self, tokenizer_path):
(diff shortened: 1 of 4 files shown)


nsorros:
Note that the flake8 error is fixed in #313

wellcomeml/ml/transformers_tokenizer.py
```diff
@@ -99,14 +100,19 @@ def tokenize(self, text):
             raise NotImplementedError
 
     def encode(self, text):
-        if type(text) == list:
+        if type(text) in [list, np.ndarray]:
```
aCampello, line 103: I think people usually prefer `isinstance` for these comparisons because then it also holds for sub-classes. 

Also, maybe you only need it to be an iterable of any sort?

aCampello (approved):
Nice - looking forward to seeing SHAP.

wellcomeml/ml/transformers_tokenizer.py
nsorros: I somehow do not use `isinstance` as often as I should. In this case though can `list` or `np.ndarray` be subleased in a meaningful way?

How would you check for an iterable? `hasattr(text, __iter__)`?

aCampello: Oh sorry, I missed this comment. Yeah, I don't think there are any sub-classes of list or Numpy arrays that I can recall of. It's a general practice, in other repositories that I used typed python and `mypy` checks, it fails to identify `type(x) = ` kind of comparisons.

I used this to check for iterables before (https://docs.python.org/3/library/typing.html#user-defined-generic-types):
```python
from collections.abc import Iterable
isinstance(x, Iterable)
```

PR #318, opened 2021-07-15
Implement multilabel attention in CNN

nsorros:
Description
---

The idea is that each label will have a separate self attention mechanism that will attend to the inputs differently. This sounds like an expensive operation but it is quite cheap as each attention requires a vector of the same dimension as the words for example `400`. Of course we need 30k of those but this not different to the dimensionality of the embedding layer which has a vocabulary size of 30k as well.

See a schematic of multi-label attention from this [paper](https://papers.nips.cc/paper/2019/file/9e6a921fbc428b5638b3986e365d4f21-Paper.pdf) below

Screenshot 2021-07-28 at 12 54 48 PM

Checklist
---

- [x] Added tests

### DIFF
--- a/wellcomeml/ml/cnn.py
+++ b/wellcomeml/ml/cnn.py
@@ -222,6 +222,9 @@ class CNNClassifier(BaseEstimator, ClassifierMixin):
             x = tf.keras.backend.sum(x, axis=1)
         elif self.feature_approach == 'concat':
             x = tf.keras.layers.Flatten()(x)
+        elif self.feature_approach == 'multilabel-attention':
+            # in multilabel attention each head represents an output
+            pass
         else:
             raise NotImplementedError
         x = tf.keras.layers.Dense(
@@ -233,11 +236,17 @@ class CNNClassifier(BaseEstimator, ClassifierMixin):
         output_activation = (
             "sigmoid" if nb_outputs == 1 or self.multilabel else "softmax"
         )
+        if self.feature_approach == "multilabel-attention":
+            # each output comes from an attention head
+            nb_outputs = 1
         out = tf.keras.layers.Dense(
             nb_outputs,
             activation=output_activation,
             kernel_regularizer=l2,
         )(x)
+        if self.feature_approach == "multilabel-attention":
+            # out shape is (batch_size, attention heads (or outputs), 1)
+            out = tf.keras.layers.Flatten()(out)
         model = tf.keras.Model(inp, out)
 
         learning_rate = tf.keras.optimizers.schedules.ExponentialDecay(
--- a/wellcomeml/ml/cnn.py
+++ b/wellcomeml/ml/cnn.py
@@ -177,10 +177,14 @@ class CNNClassifier(BaseEstimator, ClassifierMixin):
             return tf.keras.layers.add([x1, x2])
 
         def residual_attention(x1):
-            x2 = HierarchicalAttention(self.attention_heads)(x1)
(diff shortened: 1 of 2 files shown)


nsorros:
This is also ready for review @aCampello

nsorros:
This PR would love some feedback 😬 @aCampello

nsorros:
The last commit did not seem to have triggered the tests 🤔

nsorros:
This waits for a wonderful review @aCampello

aCampello:
Apologies for taking embarrassingly long, will give it a go today.

wellcomeml/ml/cnn.py
```diff
@@ -233,11 +240,17 @@ def residual_attention(x1):
         output_activation = (
             "sigmoid" if nb_outputs == 1 or self.multilabel else "softmax"
         )
+        if self.feature_approach == "multilabel-attention":
+            # each output comes from an attention head
+            nb_outputs = 1
         out = tf.keras.layers.Dense(
             nb_outputs,
```
aCampello, line 247: I don't understand, if this is multilabel attention, the dense layer will only have one unit?

aCampello: Ah, I think I worked out the maths. The layer will be `(batch_size, number_of_labels, 1)`

wellcomeml/ml/cnn.py
```diff
@@ -233,11 +240,17 @@ def residual_attention(x1):
         output_activation = (
             "sigmoid" if nb_outputs == 1 or self.multilabel else "softmax"
         )
+        if self.feature_approach == "multilabel-attention":
+            # each output comes from an attention head
+            nb_outputs = 1
```
aCampello, line 245: It's a bit weird that you use nb_outputs as it's actual value for the output activation, then you change it then use it again, but that's just style

aCampello (approved):
I really apologise for embarrassingly dropping the ball here. I was supposed to review this change while you were on holiday, but the business got in the way.


Notice that in the [paper](https://papers.nips.cc/paper/2019/file/9e6a921fbc428b5638b3986e365d4f21-Paper.pdf), it seems that the model is _learning_ the hierarchy of labels.

Now in your case you already have the tree structure - it comes from MeSH! It seems that there should be way of leveraging that instead.

wellcomeml/ml/cnn.py
nsorros: 👌

wellcomeml/ml/cnn.py
nsorros: I understand this is confusing, will try to improve terminology in the future

nsorros:
> I really apologise for embarrassingly dropping the ball here. I was supposed to review this change while you were on holiday, but the business got in the way.
> 
> Notice that in the [paper](https://papers.nips.cc/paper/2019/file/9e6a921fbc428b5638b3986e365d4f21-Paper.pdf), it seems that the model is _learning_ the hierarchy of labels.
> 
> Now in your case you already have the tree structure - it comes from MeSH! It seems that there should be way of leveraging that instead.

Indeed they use a tree but not from MeSH. Constructing that tree might be something we will have to look soon but the idea behind it is described here http://cs.uef.fi/sipu/pub/Balanced_kmeans_ssspr14.pdf

aCampello:
Feel free to merge. I think the new numpy causes the pipeline to fail, we'll see.

nsorros:
This is ready to merged, I mean psychologically ready

PR #327, opened 2021-07-21
Add tensorboard_log_path and verbose param to CNN

### LINKED ISSUE #85: VotingClassifier that accepts pretrained models and works for multilabel
Current implementation of VotingClassifier https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.VotingClassifier.html does not support pretrained models and does not work for multilabel.

We would like to add both these features, either to the existing implementation or provide a striped down version that supports those features like https://gist.github.com/tomquisel/a421235422fdf6b51ec2ccc5e3dee1b4

### LINKED ISSUE #328: CNN logs in Sagemaker are too verbose
The logs that the sagemaker job writes for training the CNN
neural net are dominated by the output of the fit command.

In the non log format fit writes a progress bar that is updated
after every batch update. In the log format this becomes one line
per update in a format that is hard to read. As per the documentation
in Keras https://www.tensorflow.org/api_docs/python/tf/keras/Model#fit
there is a verbose parameter that controls that. The default value
of 1 outputs the classic progress bar while a value of 2 is less verbose
and more suitable for logs, it outputs 1 line per epoch.

nsorros:
Description
---

Add two params in CNN
* verbose
* tensorboard_log_path to control where logs are written

For more information on verbose see #328 

Tensorboard logs were written by default to "logs" in working
directory where the code was running. SageMaker provides functionality
for copying those logs to s3 which can allow us to monitor progress
locally by running tensorboard and point to the s3 path. For this functionality
to work we need to be able where logs are written so that sagemaker can
pick them up.

Fixes #328 

Checklist
---

- [ ] Added tests

### DIFF
--- a/wellcomeml/ml/bert_semantic_equivalence.py
+++ b/wellcomeml/ml/bert_semantic_equivalence.py
@@ -11,7 +11,7 @@ from sklearn.model_selection import train_test_split
 from sklearn.utils.validation import check_is_fitted
 from sklearn.exceptions import NotFittedError
 
-from wellcomeml.ml.keras_utils import CategoricalMetrics, MetricMiniBatchHistory
+from wellcomeml.ml.keras_utils import CategoricalMetrics  # , MetricMiniBatchHistory
 from wellcomeml.logger import LOGGING_LEVEL, build_logger
 from wellcomeml.utils import throw_extra_import_message
 
--- a/wellcomeml/ml/bert_semantic_equivalence.py
+++ b/wellcomeml/ml/bert_semantic_equivalence.py
@@ -20,12 +20,6 @@ try:
 except ImportError as e:
     throw_extra_import_message(error=e, required_module='tensorflow', extra='tensorflow')
 
-TENSORBOARD_LOG_DIR = "logs/scalar/" + datetime.now().strftime("%Y%m%d-%H%M%S")
-CALLBACK_DICT = {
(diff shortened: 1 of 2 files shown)


aCampello (approved):
Looks good to me. Note that probably for consistency you should also apply that to:

- `ml/bilstm.py`
- `ml/bert_semantic_equivalence.py`

But if you want to merge this one first feel free to create an issue for the above.

nsorros:
> Looks good to me. Note that probably for consistency you should also apply that to:
> 
> * `ml/bilstm.py`
> * `ml/bert_semantic_equivalence.py`
> 
> But if you want to merge this one first feel free to create an issue for the above.

I am planning to refactor bilstm and combine it with CNN, maybe a good thing to do now.

Which param would you want for `bert_semantic_equivalence?` both?

wellcomeml/ml/bert_semantic_equivalence.py
```diff
@@ -213,8 +210,16 @@ def fit(self, X, y, random_state=None, epochs=3, metrics=[], **kwargs):
             self.train_steps = math.ceil(len(X_train)/self.batch_size)
             self.valid_steps = math.ceil(len(X_valid)/self.eval_batch_size)
         finally:
-            callback_objs = [CALLBACK_DICT[c] for c in self.callbacks]
-
+            callback_objs = [
+                # Issue #187
+                # MetricMiniBatchHistory()
```
aCampello, line 215: Sounds good...

PR #333, opened 2021-07-22
Refactor CNN and BiLSTM to BaseNN

### LINKED ISSUE #119: BiLSTM and CNN can be refactored
Both implementations are heavily inspired from spacy and have the same philosophy embed  encode attend predict, keeping the same dimensions along layers, using residuals to help training. The only difference is in the encode stage where one architecture is making use of a two layers bilstm as default and the other a layer cnn, this can easily change with a parameter. We can possibly name the new architecture WellcomeTextClassifier or WellcomeNeuralClassifier.

nsorros:
Description
---

This PR creates a base class for CNN and BiLSTM
and differentiates them with a parameter feature_encoder
as this is the only difference between them i.e. how
we encode the tokens to context aware embeddings

Fixes #119

Checklist
---

- [x] Added tests


aCampello:
I think this is a great idea, but there seems to be tests that take 1000 epochs, so we need to fix that before merging.

nsorros:
> I think this is a great idea, but there seems to be tests that take 1000 epochs, so we need to fix that before merging.

Those are to check for early stopping which seems broken, will come back to it and finish this

nsorros:
Closing this for now but we can reopen when this work resumes

PR #341, opened 2021-07-27
Add tokenize to KerasTokenizer

### LINKED ISSUE #340: KerasTokenizer does not implement tokenize
`TransformersTokenizer` implements a `tokenize` method which is needed for expanations.
Unfortunately during #315 I forgot to implement it so it is missing.

nsorros:
Description
---

As explained in #340, i forgot to add a tokenize method in `KerasTokenizer` during #315. As
I am using `TransformersTokenizer` in the latest models, this was missed but this PR adds it
for consistency and so that `KerasTokenizer` works with `SHAP`

Fixes #340

Checklist
---

- [x] Added tests

### DIFF
--- a/wellcomeml/ml/keras_vectorizer.py
+++ b/wellcomeml/ml/keras_vectorizer.py
@@ -47,6 +47,12 @@ class KerasTokenizer():
             return self.tokenizer.sequences_to_texts([encoded_text])[0]
         return self.tokenizer.sequences_to_texts(encoded_text)
 
+    def tokenize(self, text):
+        if type(text) == str:
+            return text.split()
+        else:
+            return [t.split() for t in text]
+
 
 class KerasVectorizer(BaseEstimator, TransformerMixin):
     def __init__(self, vocab_size=None, sequence_length=None, oov_token="<OOV>",


wellcomeml/ml/keras_vectorizer.py
```diff
@@ -47,6 +47,12 @@ def decode(self, encoded_text):
             return self.tokenizer.sequences_to_texts([encoded_text])[0]
         return self.tokenizer.sequences_to_texts(encoded_text)
 
+    def tokenize(self, text):
```
aCampello, line 50: Is that correct? Is that really the type of tokenisation we wanna do? (Just split by spaces?)

It is slightly different from transformers (when you call `.encode` and get the tokens), which I think also maybe excludes some characters etc...

nsorros: We want to expose the tokenisation that Keras uses see https://www.tensorflow.org/api_docs/python/tf/keras/preprocessing/text/Tokenizer. They seem to split by whitespace `' '`.

In transformers we similarly expose the `tokenize` method they have but in keras there is no method so we are just recreating it.

aCampello (approved):

PR #346, opened 2021-07-29
Fix/tf idf fit transform

### LINKED ISSUE #345: Tf-idf vectoriser is lemmatising twice
The transformations should only be in "transform" not in "fit".

Atm we have:

```bash
2021-07-29 13:02:16 wellcomeml.ml.frequency_vectorizer INFO: Using spacy pre-trained lemmatiser.
2021-07-29 13:03:15 wellcomeml.ml.frequency_vectorizer INFO: Fitting vectorizer.
2021-07-29 13:03:24 wellcomeml.ml.frequency_vectorizer INFO: Using spacy pre-trained lemmatiser.
```

When we run fit-transform this is causing problems because it's double lemmatising.

aCampello:
Description
---

Fixes #345 

Also improves performance a bit by initialising spacy open initialisation of class.

Checklist
---

- [ ] Added tests

### DIFF
--- a/wellcomeml/ml/frequency_vectorizer.py
+++ b/wellcomeml/ml/frequency_vectorizer.py
@@ -35,7 +35,7 @@ class WellcomeTfidf(TfidfVectorizer):
         """
         self.embedding = "tf-idf"
         self.use_regex = use_regex
-        self.use_spacy_lemmatizer= use_spacy_lemmatizer
+        self.use_spacy_lemmatizer = use_spacy_lemmatizer
 
         logger.info("Initialising frequency vectorizer.")
 
@@ -130,4 +130,3 @@ class WellcomeTfidf(TfidfVectorizer):
         X = self._pre_transform(X)
 
         return super().fit_transform(X, y=y)
-
--- a/wellcomeml/ml/frequency_vectorizer.py
+++ b/wellcomeml/ml/frequency_vectorizer.py
@@ -41,6 +41,11 @@ class WellcomeTfidf(TfidfVectorizer):
 
         super().__init__(**kwargs)
 
+        self.nlp = spacy.blank("en")
+        self.nlp.add_pipe("lemmatizer", config={"mode": "lookup"})
+        self.nlp.initialize()
+
+
     @classmethod
     def save_transformed(cls, path, X_transformed):
         """Saves transformed embedded vectors"""
@@ -84,10 +89,6 @@ class WellcomeTfidf(TfidfVectorizer):
 
         """
 
-        nlp = spacy.blank("en")
-        nlp.add_pipe("lemmatizer", config={"mode": "lookup"})
-        nlp.initialize()
-
         logger.info("Using spacy pre-trained lemmatiser.")
         if remove_stopwords_and_punct:
             return [
@@ -98,14 +99,15 @@ class WellcomeTfidf(TfidfVectorizer):
                     and not token.is_punct
                     and token.lemma_ != "-PRON-"
                 ]
-                for doc in nlp.pipe(X)
+                for doc in self.nlp.pipe(X)
             ]
         else:
             return [
-                [token.lemma_.lower() for token in doc] for doc in nlp.pipe(X)
+                [token.lemma_.lower() for token in doc] for doc in self.nlp.pipe(X)
             ]
 
     def transform(self, X, regex=True, spacy_lemmatizer=True, *_):
+        print("Ba")
         if regex:
             X = self.regex_transform(X)
         if spacy_lemmatizer:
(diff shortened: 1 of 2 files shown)


nsorros:
It would be great to add a test to demonstrate the fix i.e. a fix that would fail prior to fixing this

wellcomeml/ml/frequency_vectorizer.py
```diff
@@ -115,16 +115,9 @@ def transform(self, X, regex=True, spacy_lemmatizer=True, *_):
...
-
-        X = [" ".join(text) for text in X]
-
-        super().fit(X)
-
+    def fit(self, X, *_):
+        super().fit()
         return self
+
+    def fit_transform(self, X, y=None):
```
nsorros, line 122: isn't this provided for free from sklearn?

nsorros:
How does this prevent the `lemmatiser` running twice?

aCampello:
> How does this prevent the `lemmatiser` running twice? Is it because it does not run during `fit`? If so then that sounds wrong

The fit function of TfIdf is stateless, it does nothing. All the heavy-lifting is done in `.transform`. So any transformations should only run in `.transform`, but we were wrongly putting it into `.fix`. This doesn't interfere that much, but it was causing weird errors like, if you run `X_transform = vectorizer.fit_transform(X)` and then `X_transform_2 = vectoriser.fit(X)` you wouldn't end up with the same results. 

I will add a test to demonstrate that and also fix the test.

aCampello:
Check: 

https://github.com/scikit-learn/scikit-learn/blob/2beed55847ee70d363bdbfe14ee4401438fba057/sklearn/feature_extraction/text.py#L720-L768

aCampello:
Thanks for the test suggestion @nsorros . 

So, I added a test to check that `.fit(X).transform(X)` was the same as `fit_transform(X)`. Without the new code it fails:

```bash
        # Asserts that the result of transform is almost the same as fit transform
>       assert (X_embed-X_embed_2).sum() == pytest.approx(0, abs=1e-6)
E       assert 0.18267581368159946 == 0 ± 1.0e-06
E         +0.18267581368159946
E         -0 ± 1.0e-06

tests/test_frequency_vectorizer.py:47: AssertionError
```

With the new code it passes.

Check https://github.com/wellcometrust/WellcomeML/pull/346/commits/d272de193a79aec71525e5be312f8315d55fa3bb

wellcomeml/ml/frequency_vectorizer.py
aCampello: That's why the `TfId` class in sklearn has to implement their own `.fit.` even though it's a transform mixin.

nsorros: 👍

nsorros:
> Check:
> 
> https://github.com/scikit-learn/scikit-learn/blob/2beed55847ee70d363bdbfe14ee4401438fba057/sklearn/feature_extraction/text.py#L720-L768

This is the `fit` of `HashingVectorizer` which is indeed stateless. TfidfVectorizer is not `stateless`, it updates the vocabulary see https://github.com/scikit-learn/scikit-learn/blob/2beed55847ee70d363bdbfe14ee4401438fba057/sklearn/feature_extraction/text.py#L1806

nsorros:
> > How does this prevent the `lemmatiser` running twice? Is it because it does not run during `fit`? If so then that sounds wrong
> 
> The fit function of TfIdf is stateless, it does nothing. All the heavy-lifting is done in `.transform`. So any transformations should only run in `.transform`, but we were wrongly putting it into `.fix` too. This doesn't interfere that much because "doubly" lemmatising something should be almost idempotent, but it was causing weird errors like, if you run `X_transform = vectorizer.fit_transform(X)` and then `X_transform_2 = vectorizer.fit(X)` you wouldn't end up with the same results.
> 
> I will add a test to demonstrate that and also fix the test.

I think this is wrong. `fit` calculates the vocabulary and you were correctly lemmatising before doing so, you should lemmatise both for fit and transform as they do different things, kind of like preprocessing, you need to preprocess with the same function.

I think what was potentially going wrong is that `fit_transform` passes the X to both fit and transform but you were modifying X in `fit`, you need to save in a different variable before calling the `super().fit`

nsorros (changes requested):
I think introduces a problem as the vocabulary is calculated on the non lemmatised text whereas transform lemmatises. This can lead to unexpected results.

aCampello:
> > Check:
> > https://github.com/scikit-learn/scikit-learn/blob/2beed55847ee70d363bdbfe14ee4401438fba057/sklearn/feature_extraction/text.py#L720-L768
> 
> This is the `fit` of `HashingVectorizer` which is indeed stateless. TfidfVectorizer is not `stateless`, it updates the vocabulary see https://github.com/scikit-learn/scikit-learn/blob/2beed55847ee70d363bdbfe14ee4401438fba057/sklearn/feature_extraction/text.py#L1806

You're right, the link was wrong. The TfIdf class is actually very clunky, all the heavy-lifting is done in `fit_transform` and then the `fit` just calls `fit_transform`. 

I'll modify it to look more like TfIdf, thanks for the review.

nsorros (approved):
LGTM 👍

aCampello:
Ok, merging this guy, thanks for spotting the error @nsorros !

Issue #310, opened 2021-07-07
Codecov is broken


aCampello:
Codecov is broken in `main` and in the PRs.

`Unable to find commit in Github` shows in codecov.io

aCampello:
This is now fixed - it was some intermittent codecov issue.

Issue #251, opened 2021-03-19
Codecov not working sometimes


nsorros:
This is a spin off from #246 where I noticed that codecov was not working. After some investigation I realised that this is not a new behaviour and there is an error that pops up in the interface see the image. Another PR with no codecov is #242 but you can find most of them by following the PR with the error from codecov.

![Screenshot 2021-03-19 at 1 41 07 PM](https://user-images.githubusercontent.com/4975761/111775434-6af99180-88b9-11eb-8b22-fe6a7f5b676a.png)

nsorros:
FYI @aCampello @lizgzil

aCampello:
According to the docs

```
In order to produce an accurate coverage comparison, Codecov must have the coverage report for the pull request base.

Best Practice: Always fork a branch from a successful CI commit that has coverage uploaded, as illustrated below.

This issue occurs when the base commit (the parent commit of the first commit on the pull request) did not upload coverage and/or failed CI tests.
```

I have no idea how this might be happening since we always compare to main?

aCampello:
https://docs.codecov.io/docs/error-reference

nsorros:
I think there is some conversation about it happening when commits change i.e. force pushing or squash etc which makes sense according to what you found. In my PR codecov failed before because I had to fix the command that runned it so that explains it as well. So let's monitor this, maybe it is not a bug but a feature. If that is the source I wonder whether there is anything we can do to fix it when it occurs.

nsorros:
I will investigate this even though it might not help with hacktoberfest.

nsorros:
Maybe #365 is not needed, we could reason about this without another PR.

As we can see in #365 the codecov failed. if we go to codecov it says missing base report.
![Screenshot 2021-10-20 at 12 32 41 PM](https://user-images.githubusercontent.com/4975761/138067949-5a3137c4-30d1-4384-99fd-b77975ab0dde.png)

The previous PR that was merged was #364, check the commit hashes
![Screenshot 2021-10-20 at 12 33 17 PM](https://user-images.githubusercontent.com/4975761/138068072-36924d6b-d3b4-4508-b525-7cdf973c2993.png)

And now check the hashes in main
![Screenshot 2021-10-20 at 12 33 52 PM](https://user-images.githubusercontent.com/4975761/138068137-5e9abd4e-4a1d-4a04-949e-850035653d5c.png)

[...]

PR #365, opened 2021-10-20
Investigate why codecov breaks

nsorros:
Description
---

This is an investigation of #310. The theory is that codecov breaks when we mess up with commit hashes somehow. One way this can happen is if a PR is behind main and we merge using rebase and squash. This would alter the commit hash without running tests or codecov so the report for main will be missing.

This PR dropped some commits from main and made a simple change. The goal is to merge with `rebase and squash` and then open another PR to test whether `codecov` is broken. If it is, I suspect that merging the new PR will fix that as the tests and codecov would run for that.

Checklist
---

- [ ] Added tests


nsorros:
Closing this, investigation details #251

PR #374, opened 2021-11-04
Release 2.0.0

### LINKED ISSUE #373: Release 2.0
Some of the hacktoberfest issues are breaking changes, so we need to call the next release 2.0

nsorros:
Description
---

Release 2.0.0 which, among other things, introduces a breaking change and makes WellcomeML light
both contributions from @acampello.

The breaking change forces users to explicitly import wellcome ml models
so instead of `from wellcomeml.ml import CNNClassifier` you now have to
do `from wellcomeml.ml.cnn import CNNClassifier`

WellcomeML light ~109MB is possible through a careful creation of extras.
In particular we introduce `core` and `transformers` as additional extras.
This allows us to create python specific implementaitons that are quick
and do not require extra "heavy" libraries but also move us towards a very
refined installation process from the user where they install what they need.
The user might even not need to install as they might have the library needed
already in their virtualenv, for example if they want to use CNNClassifier and
have `tensorflow` installed, installing just wellcomeml will work. 

Fixes #373

Checklist
---

- [x] Added link to Github issue or Notion card
- [ ] Added tests


aCampello:
That's the release we were waiting for!

aCampello:
That's the PR that we all wanted to see!!!!

aCampello (approved):

nsorros:
Released 🎉 https://pypi.org/project/wellcomeml/2.0.0/

nsorros:
`time pip install --no-cache-dir wellcomeml` runs in `34s` 🚀 and consumes only 111MB 🪶

PR #386, opened 2022-04-07
Feature/nice heatmap

aCampello:
Description
---

Creates an util to plot a beautiful and interactive co-ocurrence heatmap with the Wellcome palette and colours. Can be used for different use cases (e.g. confusion matrix, to inspect multilabel classification, to plot overlap between tags/areas), hence the idea of adding it to `wellcomeml`. The heatmap will look like the below:

bokeh_plot-15

### DIFF
--- a/examples/heatmap.py
+++ b/examples/heatmap.py
@@ -1,0 +1,18 @@
+from wellcomeml.viz.elements import plot_heatmap
+
+# Fake co_occurrence matrix of the concepts "Data Science", "Machine Learning" and "Science"
+
+co_occurrence = [
+   {"concept_1": "Data Science", "concept_2": "Machine Learning", "value": 1},
+   {"concept_1": "Machine Learning", "concept_2": "Data Science", "value": 0.3},
+   {"concept_1": "Science", "concept_2": "Data Science", "value": 1},
+   {"concept_1": "Science", "concept_2": "Machine Learning", "value": 1},
+   {"concept_1": "Data Science", "concept_2": "Science", "value": 0.05},
+   {"concept_1": "Machine Learning", "concept_2": "Science", "value": 0.01}
+]
+
+# Plot in blue
+plot_heatmap(co_occurrence, file='test-blue.html', color="Blue Lagoon")
+
+# Plot in gold
+plot_heatmap(co_occurrence, file='test-gold.html', color="Tahiti Gold")
--- a/examples/heatmap.py
+++ b/examples/heatmap.py
@@ -3,16 +3,18 @@ from wellcomeml.viz.elements import plot_heatmap
 # Fake co_occurrence matrix of the concepts "Data Science", "Machine Learning" and "Science"
 
 co_occurrence = [
-   {"concept_1": "Data Science", "concept_2": "Machine Learning", "value": 1},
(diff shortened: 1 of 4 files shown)


wellcomeml/viz/elements.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+def plot_heatmap(co_occurrence,
+                 concept_order=None,
+                 file="heatmap.html",
+                 notebook=True,
+                 title="",
+                 notebook_url='localhost:8888',
+                 color='Blue Lagoon',
+                 metadata_to_display=[]):
```
justinbt1, line 18: Noticed that there is a list as a default parameter, is this a deliberate choice? If not, maybe consider using None as the default for metadata_to_display as it should still evaluate correctly with the if statement within the function. In case anyone hasn't come across this before, this will help avoid any [unexpected mutability behaviours](https://docs.python-guide.org/writing/gotchas/).

wellcomeml/viz/elements.py
```diff
@@ -0,0 +1,113 @@
...
+        "concept_1", "concept_2", and the value of the edge, between 0 and 1.
+
+        For example:
+
+        co_ocurrence = [
+            {"concept_1": apple, "concept_2": fruit, "value": 0.2},
+            {"concept_1": apple, "concept_2": vegetable, "value": 0.1},
+            {"concept_1": banana, "concept_2": fruit, "value": 0.1}
+        ]
+
```
justinbt1, line 34: Really nice detailed description of the co_occurance data structure 😃

justinbt1 (changes requested):
The bokeh visual looks great! The code runs first time without error and the docstrings / comments are well placed and informative. Though I haven't used bokeh much, so there is a chance I missed something.

My only concern is the use of the mutable [ ] as a parameter default for the metadata_to_display parameter, see comment for details.

examples/heatmap.py
```diff
@@ -0,0 +1,18 @@
...
+   {"concept_1": "Science", "concept_2": "Machine Learning", "value": 1},
+   {"concept_1": "Data Science", "concept_2": "Science", "value": 0.05},
+   {"concept_1": "Machine Learning", "concept_2": "Science", "value": 0.01}
+]
+
+# Plot in blue
+plot_heatmap(co_occurrence, file='test-blue.html', color="Blue Lagoon")
+
+# Plot in gold
+plot_heatmap(co_occurrence, file='test-gold.html', color="Tahiti Gold")
```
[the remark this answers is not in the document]
aCampello: Good that it ran!

aCampello:
> The bokeh visual looks great! The code runs first time without error and the docstrings / comments are well placed and informative. Though I haven't used bokeh much, so there is a chance I missed something.
> 
> My only concern is the use of the mutable [ ] as a parameter default for the metadata_to_display parameter, see comment for details.

Good shout. I don't think we're modifying that in the function, but probably better to change it as you mentioned: https://github.com/wellcometrust/WellcomeML/pull/386/commits/ac6ac641a3a0ec7f1adb82966d8acbef28bf457f

justinbt1 (approved):
No further issues and code still runs fine, visuals display correctly when example code is run.

